PR #6, opened 2024-12-10
Add sign extension ci

vladavoX:


.github/workflows/sing-extension.yml
```diff
@@ -0,0 +1,65 @@
+# This workflow will perform signing of a plugin
+
+name: Sign Extension
+
+on:
+  workflow_run:
+    workflows: [ "CI Workflow" ]
+    types: [ completed ]
+    branches: [ "master" ]
```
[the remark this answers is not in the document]
ivansenic: actually let's only have a manual trigger for now

CSXS/manifest.xml
```diff
@@ -1,19 +1,24 @@
 <?xml version="1.0" encoding="UTF-8" standalone="no"?>
 <ExtensionManifest
-	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" ExtensionBundleId="com.plainlyvideos.after-effects-plugin" ExtensionBundleVersion="1.0" Version="12.0">
+	Version="11.0"
+	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" 
+	ExtensionBundleName="Plainly plugin"
```
ivansenic, line 5:
```suggestion
	ExtensionBundleName="Plainly Videos" 
```

.github/workflows/sing-extension.yml
```diff
@@ -0,0 +1,73 @@
...
+
+  workflow_dispatch:
+    inputs:
+      reason:
+        description: 'Reason for signing'
+        required: true
+
+jobs:
+  sign:
```
ivansenic, line 18: we need then a job before this one that prepares node modules for archiver, should be simple and then exchange between jobs: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/storing-and-sharing-data-from-a-workflow#passing-data-between-jobs-in-a-workflow

ivansenic (changes requested):

PR #10, opened 2024-12-16
Make plugin CSXS11 compatible

vladavoX:
- change `archiver` to `5.3.2`
- change `require("node:something")` to `require("something")`
- move bundeled script to body tag in index.html
- change lib files to CSXS11

### AE25
image

### AE24
image

### DIFF
--- a/src/lib/CEPEngine_extensions.js
+++ b/src/lib/CEPEngine_extensions.js
@@ -16,7 +16,7 @@
 //
 // Note: So far all native file i/o functions are synchronous, and aynchronous file i/o is TBD.
 
-/** Version v12.0.0 */
+/** Version v11.0.0 */
 
 /*jslint vars: true, plusplus: true, devel: true, browser: true, nomen: true, indent: 4, forin: true, maxerr: 50, regexp: true */
 /*global define, native */
--- a/src/lib/Vulcan.js
+++ b/src/lib/Vulcan.js
@@ -11,9 +11,7 @@
  *
  **************************************************************************************************/
 
-/** CEP-Vulcan.js v11.1.0 - This JS Bindings for Vulcan is compatible with Vulcan v7.0 */
-
-/**
+/** Vulcan - v11.2.0 */
 
 /**
  * @class Vulcan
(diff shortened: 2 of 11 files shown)


src/lib/CEPEngine_extensions.js
```diff
@@ -16,7 +16,7 @@
 //
 // Note: So far all native file i/o functions are synchronous, and aynchronous file i/o is TBD.
 
-/** Version v12.0.0 */
+/** Version v11.0.0 */
```
ivansenic, line 19: how can it be that in these files only a version was changed in the comment and nothing else?

src/lib/Vulcan.js
```diff
@@ -11,9 +11,7 @@
  *
  **************************************************************************************************/
 
-/** CEP-Vulcan.js v11.1.0 - This JS Bindings for Vulcan is compatible with Vulcan v7.0 */
-
-/**
+/** Vulcan - v11.2.0 */
```
ivansenic, line 14: and here we were at 11.1, not at 11.2, wth?

vite-plugins/jsBottomOfIndex.ts
```diff
@@ -0,0 +1,25 @@
...
+      const matchResult = replacedHtml.match(
+        /<script[^>]*>(.*?)<\/script[^>]*>/,
+      );
+
+      if (matchResult) {
+        const scriptTag = matchResult[0];
+        console.log('\n SCRIPT TAG', scriptTag, '\n');
+        replacedHtml = replacedHtml.replace(scriptTag, '');
+        replacedHtml = replacedHtml.replace(
+          '<!-- # INSERT SCRIPT HERE -->',
```
ivansenic, line 15: we could have done it without the marker, just search for `<body>` and replace with `<body> + <script>`

ivansenic (approved):

src/lib/Vulcan.js
vladavoX: https://github.com/Adobe-CEP/CEP-Resources/tree/master/CEP_11.x
these are the they files from CEP_11. I've ran formater (because I've originaly ran that on these files also), and this is the only change for that file

vladavoX: same for, CSInterface. not many changes. Only formating changes

PR #14, opened 2024-12-18
closes #13: Performance optimization / UX improvement

### LINKED ISSUE #13: Performance optimization / UX improvement
Big projects take some time to collect files, without showing any progress bar.

In addition, we could improve performance:
1. Make jsx file return a list of all paths to footage and font files - this could be only one function to avoid going through the whole project twice
2. Copy all these files in parallel using nodejs 
3. Make a progress bar (like 15/142 files collected) / or just an infinite spinner because it should be much faster after 1. and 2.

vladavoX:
Original issue: #13 

- Move copy file logic to node
- Add loading spinner to Button
- Fix Error catching

### DIFF
--- a/src/node/errors.ts
+++ b/src/node/errors.ts
@@ -1,0 +1,15 @@
+export class CollectFootageError extends Error {
+  errorPaths: string[];
+  constructor(errorPaths: string[]) {
+    super();
+    this.errorPaths = errorPaths;
+  }
+}
+
+export class CollectFontsError extends Error {
+  errorPaths: string[];
+  constructor(errorPaths: string[]) {
+    super();
+    this.errorPaths = errorPaths;
+  }
+}
--- a/src/node/errors.ts
+++ b/src/node/errors.ts
@@ -1,7 +1,9 @@
 export class CollectFootageError extends Error {
   errorPaths: string[];
   constructor(errorPaths: string[]) {
-    super();
+    super(
+      `Could not collect footage for the following paths: ${errorPaths.join(', ')}`,
+    );
     this.errorPaths = errorPaths;
   }
 }
@@ -9,7 +11,9 @@ export class CollectFootageError extends Error {
 export class CollectFontsError extends Error {
   errorPaths: string[];
   constructor(errorPaths: string[]) {
-    super();
+    super(
+      `Could not collect footage for the following paths: ${errorPaths.join(', ')}`,
+    );
     this.errorPaths = errorPaths;
   }
 }
--- a/src/node/export.ts
+++ b/src/node/export.ts
@@ -5,6 +5,7 @@ const path = require('path');
 
 // @ts-ignore
 import CSInterface from '../lib/CSInterface';
+import { CollectFontsError, CollectFootageError } from './errors';
 import type { Fonts, Footage, ProjectInfo } from './types';
 import {
   evalScriptAsync,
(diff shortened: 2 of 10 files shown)


src/node/export.ts
```diff
@@ -26,30 +33,77 @@ function selectFolder(callback: (result: string) => void) {
...
- * @example
- * untildify('~/Documents') // => '/Users/username/Documents'
- * @param {string} pathWithTilde - The path to transform.
- * @returns {string} The path with the tilde replaced.
- */
-function untildify(pathWithTilde: string): string {
-  if (typeof pathWithTilde !== 'string') {
-    throw new TypeError(`Expected a string, got ${typeof pathWithTilde}`);
+async function copyFonts(fonts: Fonts[], projectDir: string) {
+  if (fonts.length > 0) {
```
danixeee, line 57:
```ts
if (fonts.length === 0) {
    return;
}
```

this way you don't have to indent the rest of the code

src/node/export.ts
```diff
@@ -26,30 +33,77 @@ function selectFolder(callback: (result: string) => void) {
...
+      const src = finalizePath(font.fontLocation);
+
+      const dest = path.join(
+        fontsDir,
+        `${font.fontName}.${font.fontExtension}`,
+      );
+      return fsPromises.copyFile(src, dest);
+    });
+
+    const errors = await runInParallelReturnRejected(fontPromises);
```
danixeee, line 71: Errors here will be errors thrown by `copyFile`. So, you will get the full message like `Error: ENOENT: no such file or directory, copyfile ...`. You can wrap `return fsPromises.copyFile(src, dest);` with try-catch, and `throw Error(src);` for example

src/node/export.ts
```diff
@@ -26,30 +33,77 @@ function selectFolder(callback: (result: string) => void) {
...
+
+    if (errors.length > 0) {
+      throw new CollectFontsError(errors);
+    }
   }
+}
+
+async function copyFootage(footage: Footage[], projectDir: string) {
+  if (footage.length > 0) {
```
danixeee, line 80: same as above

src/node/export.ts
```diff
@@ -26,30 +33,77 @@ function selectFolder(callback: (result: string) => void) {
...
+      if (folder === 'Root') {
+        const dest = path.join(footageDir, footageName);
+        return fsPromises.copyFile(finalizePath(footageItem.itemName), dest);
+      }
+
+      const replaced = folder.replace('Root', '');
+      generateFolders(path.join(footageDir, replaced));
+
+      const dest = path.join(footageDir, replaced, footageName);
+      return fsPromises.copyFile(src, dest);
```
danixeee, line 98: same as above, wrap with try-catch and throw only src

src/node/errors.ts
danixeee, line 1: So, you are not using `errorPath` in the `ExportForm.tsx` now, and these error messages are empty. Meaning that user will get a blank error notification. For now, I think you can just populate the error messages here and later we will see.

e.g.

```
export class CollectFootageError extends Error {
  errorPaths: string[];
  constructor(errorPaths: string[]) {
    super("Could not collect footage for the following paths: " + errorPaths.join(", "));
    this.errorPaths = errorPaths;
  }
}
```

danixeee: You can try to simulate the error of copyFile and see how it looks

danixeee (approved):

PR #16, opened 2024-12-19
Add api key to settings

vladavoX:


src/hooks/useNotification.ts
```diff
@@ -0,0 +1,47 @@
...
+      setNotification({
+        title,
+        type: NotificationType.ERROR,
+        description,
+      });
+
+      setTimeout(() => {
+        clearNotification();
+      }, 3000);
+    },
```
vladavoX, line 37: remove timeout from error notifications

vladavoX:
image

image

src/node/settings.ts
```diff
@@ -0,0 +1,77 @@
...
+    await get('/api/v2/integrations/appmixer/user-profile', apiKey);
+  } catch (error) {
+    throw new Error('Invalid API key');
+  }
+
+  const settings = await getSettings();
+  const newSettings = { ...settings, apiKey, hasPin: false };
+
+  if (pin) {
+    const secret = `${pin.first}${pin.second}${pin.third}${pin.fourth}`;
```
ivansenic, line 39: can pin be a class that can return the pin and avoid doing this everywhere?

src/node/settings.ts
```diff
@@ -0,0 +1,77 @@
...
+import { encode } from './encoding';
+import { get } from './request';
+import type { Settings } from './types';
+
+const device = os.type();
+
+async function getSettings(): Promise<Settings> {
+  let settingsFile: string;
+
+  if (device === 'Windows_NT') {
```
ivansenic, line 17: os switch should be in `retrieveSettingsFile()` imo

ivansenic (changes requested):
Could not review as it does not work entering the API key

.env.development
```diff
@@ -1 +1,2 @@
-VITE_APP_PLAINLY_ENV=development
\ No newline at end of file
+VITE_APP_PLAINLY_ENV=development
+VITE_API_BASE_URL=api.test.plainlyvideos.com
```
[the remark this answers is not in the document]
vladavoX: I will do this later please, because I don't want to run my FE-BE to be able to test, much easier with app.test

package.json
```diff
@@ -5,7 +5,7 @@
   "type": "module",
   "scripts": {
     "dev": "NODE_ENV=development node ./script/updateManifest.cjs --dev && vite build --mode development --watch",
-    "build-test": "NODE_ENV=test node ./script/updateManifest.cjs && tsc -b && vite build && webpack --mode production",
+    "build-test": "NODE_ENV=production node ./script/updateManifest.cjs && tsc -b && vite build --mode test && webpack --mode production",
```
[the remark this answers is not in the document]
vladavoX: second one is a must, becuse `vite` takes that env from its `mode`, first one should be good now, because we want manifest to be same as production?

src/node/types/index.ts
```diff
@@ -14,3 +14,8 @@ export interface Footage {
   itemName: string;
   itemFolder: string;
 }
+
+export interface Settings {
+  apiKey?: string;
+  hasPin?: boolean;
+}
```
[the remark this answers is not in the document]
vladavoX: I am not sure that it makes even sense that inside .ini file we make object strucutre;

because I think ini file is similar to env file for example
key=value
anotherKey=anotherValue

But I did it this way because of `JSON.parse`, or do we even care, and make it look the way we want?

src/node/settings.ts
vladavoX: not done yet

.env.development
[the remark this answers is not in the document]
ivansenic: Doesn't matter, you have 2 options now:

* `build-test`
* overwrite locally why you develop

committed code should be correct

src/components/settings/SettingsForm.tsx
```diff
@@ -0,0 +1,316 @@
...
+              <Label label="PIN (recommended)" htmlFor="pin" />
+              <Description>
+                Setup your four digit PIN to encrypt your API key stored on your
+                device. You will be asked to enter this PIN on first request
+                sent to Plainly every time you restart the extension. If you
+                don&apos;t want to use a PIN, you can leave this field empty.
+              </Description>
+            </div>
+            <div className="mt-2 grid grid-cols-6 items-center justify-items-center gap-y-2">
+              <PinInput
```
ivansenic, line 162: * please show me the pin here always
* and return that confirmation text, as I know don't know what it means

image

src/components/settings/SettingsForm.tsx
```diff
@@ -0,0 +1,316 @@
...
+  showPin?: boolean;
+  className?: string;
+}) {
+  // Create refs for each input
+  const firstInputRef = useRef<HTMLInputElement>(null);
+  const secondInputRef = useRef<HTMLInputElement>(null);
+  const thirdInputRef = useRef<HTMLInputElement>(null);
+  const fourthInputRef = useRef<HTMLInputElement>(null);
+
+  const changeDigit = useCallback(
```
ivansenic, line 231: seems not enough, if there is any input I don't have all text selected:

image

src/components/settings/SettingsForm.tsx
```diff
@@ -0,0 +1,316 @@
...
+import { getSettings } from '../../node/settings';
+import type { Pin } from '../../types';
+import { handleLinkClick } from '../../utils';
+import Button from '../common/Button';
+import Notification from '../common/Notification';
+import Description from '../typography/Description';
+import Label from '../typography/Label';
+import PageHeading from '../typography/PageHeading';
+
+export default function SettingsForm() {
```
ivansenic, line 17: * If api key is set, you could not display form, but have only an option to change API key
   * So display information only API key set + is using PIN
   * and option to edit
* Once I want to edit the key display the form but without any existing api key
  * now for example you even show the encrypted API key, which makes no sense at all

image
 


image

src/components/settings/SettingsForm.tsx
```diff
@@ -0,0 +1,316 @@
...
+      } catch (error) {
+        notifyError('Failed to save settings', (error as Error).message);
+        setLoading(false);
+        setPin(undefined);
+        setConfirmPin(undefined);
+      }
+    }
+  };
+
+  useEffect(() => {
```
ivansenic, line 65: `useSettings`

src/components/settings/SettingsForm.tsx
```diff
@@ -0,0 +1,273 @@
...
+    <form className="space-y-6 w-full text-white" onSubmit={handleSubmit}>
+      <div className="space-y-6 border-b border-white/10 pb-6">
+        <div>
+          <PageHeading heading="Settings" />
+          <Description className="mt-1">
+            Lorem ipsum dolor sit amet consectetur adipisicing elit. Deserunt
+            odio totam incidunt ratione vero ex possimus cumque voluptatibus
+            ipsa commodi doloribus dolorum ipsum, non aliquam earum aspernatur
+            amet ab praesentium.
+          </Description>
```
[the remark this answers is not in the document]
ivansenic: something short

src/node/settings.ts
```diff
@@ -0,0 +1,72 @@
+const fsPromises = require('fs/promises');
+const fs = require('fs');
+const path = require('path');
+
+import type { Pin } from '../types';
+import { getSettingsDirectory } from './constants';
```
ivansenic, line 6: add `getSettingsFile` as well

src/node/settings.ts
```diff
@@ -0,0 +1,72 @@
...
+async function saveSettings(settings: Settings) {
+  const settingsFile = JSON.stringify(settings, null, 2);
+  const dest = getSettingsDirectory();
+
+  await fsPromises.writeFile(
+    path.join(dest, 'settings.ini'),
+    settingsFile,
+    'utf-8',
+  );
+}
```
ivansenic, line 72: not done:

https://github.com/plainly-videos/after-effects-plugin/pull/16#discussion_r1893922462

ivansenic (changes requested):

src/components/settings/SettingsForm.tsx
```diff
@@ -0,0 +1,316 @@
...
+  const handleSubmit = async (e: React.FormEvent) => {
+    e.preventDefault();
+    setLoading(true);
+
+    if (apiKey) {
+      if (pin && Object.values(pin).some((value) => !value)) {
+        notifyError('All digits must be filled');
+        setLoading(false);
+        return;
+      }
```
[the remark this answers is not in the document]
vladavoX: it is not required, because it can be optional. This is only, if you start typing it, and leave some blank for some reason

src/components/settings/SettingsForm.tsx
```diff
@@ -0,0 +1,316 @@
...
+    const getApiKey = async () => {
+      const settings = await getSettings();
+      let apiKey = settings.apiKey;
+
+      if (apiKey && settings.hasPin && getItem('pin')) {
+        const pin = getItem('pin') as string;
+        const parsedPin = JSON.parse(pin);
+        const secret = `${parsedPin.first}${parsedPin.second}${parsedPin.third}${parsedPin.fourth}`;
+
+        apiKey = decode(secret, apiKey);
```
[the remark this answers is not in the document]
vladavoX: to display to a user if he has it already setup, to be able to see it, if he wants. But if we are going with that approach above, to not show it at all when it is already setup, then this will be removed

this comment: https://github.com/plainly-videos/after-effects-plugin/pull/16#discussion_r1895561336

src/node/constants.ts
```diff
@@ -0,0 +1,32 @@
...
+  'AppData',
+  'Roaming',
+  'Adobe',
+  'Plainly Videos',
+);
+
+const device: string = os.platform();
+
+export function getSettingsDirectory(): string {
+  if (device === 'win32') return windowsDest;
```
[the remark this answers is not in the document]
vladavoX: os.platform should return that for all windowses, as I've seen in docs

src/components/settings/SettingsForm.tsx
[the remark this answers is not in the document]
ivansenic: Yup

src/components/settings/SettingsForm.tsx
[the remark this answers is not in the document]
ivansenic: Yes correct, thanks.. again handle with class `isFilled()` method for example

src/node/constants.ts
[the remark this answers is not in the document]
ivansenic: Good, resolving

src/node/constants.ts
```diff
@@ -0,0 +1,32 @@
...
+
+  return macDest;
```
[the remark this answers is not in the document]
vladavoX: `This else clause can be omitted because previous branches break early.`
Which makes sense, becuase it must go under it, no need for `else`. Do you want me to disable this error message?

src/App.tsx
```diff
@@ -15,21 +19,35 @@ export default function App() {
   const children = useMemo(() => {
     if (currentPage === '/export') return <ExportRoute />;
     if (currentPage === '/about') return <AboutRoute />;
+    if (currentPage === '/settings') return <SettingsRoute />;
 
     return null;
   }, [currentPage]);
 
+  const [showPinOverlay, setShowPinOverlay] = useState(false);
```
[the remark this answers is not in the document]
vladavoX: removed this from this PR

src/node/settings.ts
vladavoX: are you sure? I've set the destination to be most random (Desktop), that doesn't have any of these folders and settings.json in it, and it created all for me

src/node/settings.ts
vladavoX: this is not needed now, right? or correct me if am wrong

src/App.tsx
```diff
@@ -25,7 +27,7 @@ export default function App() {
         <Button
           secondary
           onClick={reloadExtension}
-          className="absolute top-3 right-3 cursor-pointer z-30"
+          className="absolute top-3 right-3 cursor-pointer z-40"
```
vladavoX, line 30: change because of future overlay, just leave this change as it is, so I don't need to think in future again

src/hooks/useSettingsReducer.ts
```diff
@@ -0,0 +1,90 @@
+import { useCallback, useEffect, useReducer, useState } from 'react';
```
ivansenic, line 1: just `useSeetings` both file name and name of the funcion

src/node/request.ts
```diff
@@ -0,0 +1,74 @@
...
+const http = require('http');
+
+import type { ClientRequest, RequestOptions, ServerResponse } from 'http';
+import { isDev } from '../env';
+
+const protocol = isDev ? http : https;
+
+const options = {
+  hostname: import.meta.env.VITE_API_BASE_URL,
+  port: isDev ? 8080 : undefined,
```
ivansenic, line 11: move to `import.meta.env.VITE_API_BASE_PORT`

src/node/request.ts
```diff
@@ -0,0 +1,74 @@
...
+        // Resolve or Reject the Promise when the response ends
+        res.on('end', () => {
+          const { statusCode } = res;
+          if (statusCode >= 200 && statusCode < 300) {
+            try {
+              const response = JSON.parse(data);
+              resolve(response);
+            } catch (error) {
+              reject({
+                statusCode: 500,
```
ivansenic, line 36: wrong, you can not send 500 back.. reject without status code, make it optional

src/node/request.ts
```diff
@@ -0,0 +1,74 @@
...
+
+        // Collect the response data
+        res.on('data', (chunk) => {
+          data += chunk;
+        });
+
+        // Resolve or Reject the Promise when the response ends
+        res.on('end', () => {
+          const { statusCode } = res;
+          if (statusCode >= 200 && statusCode < 300) {
```
ivansenic, line 30: * why less than 300? 3xx should be fine or?
* are requests auto forwarded?
* imo `statusCode < 400` should be fine here

src/node/request.ts
```diff
@@ -0,0 +1,74 @@
...
+            } catch (error) {
+              reject({
+                statusCode: 500,
+                message: `Failed to parse response: ${(error as Error).message}`,
+              });
+            }
+          } else {
+            const error = {
+              statusCode,
+              message: data,
```
ivansenic, line 43: why are we also not parsing the data here, should also be a JSON in case of an error?

imo:

1. parse data in try
2. depending on the status code reject or approve
3. if parsing fails do same as now

src/node/settings.ts
```diff
@@ -0,0 +1,63 @@
...
+      },
+    };
+  }
+
+  try {
+    await saveSettings(newSettings);
+  } catch (error) {
+    console.log(error);
+  }
+}
```
ivansenic, line 25: I don't understand what's going on here? what's this gymnastics with `newSettings` and why?

src/node/settings.ts
```diff
@@ -0,0 +1,63 @@
...
+
+async function saveSettings(settings: Settings) {
+  const settingsContent = JSON.stringify(settings, null, 2);
+  const dest = getSettingsDirectory();
+
+  await fsPromises.writeFile(
+    path.join(dest, 'settings.json'),
+    settingsContent,
+    'utf-8',
+  );
```
ivansenic, line 62: Again, this is not done, third comment already:

1. extract settings file path to const
2. in the read settings:
   * if file exists read, parse on error default
   * if file does not exists, default
3. in write settings, ensure settings directory is created and write

src/node/constants.ts
```diff
@@ -0,0 +1,38 @@
...
+
+export function getSettingsDirectory(): string {
+  if (device === 'win32') {
+    return windowsDest;
+  }
+  return macDest;
+}
+
+export const defaultSettings: Settings = {
+  apiKey: undefined,
```
ivansenic, line 37: * why is this needed, isn't `{}` enough?
* I also think this should be defined in the `settings.ts`

src/hooks/useSettingsReducer.ts
```diff
@@ -0,0 +1,90 @@
...
+import { get } from '../node/request';
+import { retrieveSettings, setSettings } from '../node/settings';
+import type { Settings } from '../node/types';
+import { Pin } from '../types';
+
+type Action =
+  | {
+      type: 'SET_SETTINGS';
+      payload: Settings;
+    }
```
ivansenic, line 13: why is this action needed, remove

src/hooks/useSettingsReducer.ts
```diff
@@ -0,0 +1,90 @@
...
+}
+
+export const useSettingsReducer = () => {
+  const [settings, dispatch] = useReducer(settingsReducer, defaultSettings);
+  const [loading, setLoading] = useState(true);
+
+  useEffect(() => {
+    const fetchSettings = async () => {
+      try {
+        const initialSettings = await retrieveSettings();
```
ivansenic, line 42: why can the initial settings go in the reducer initialization? that was the point imo?

src/hooks/useSettingsReducer.ts
```diff
@@ -0,0 +1,90 @@
...
+      } catch (error) {
+        throw new Error(
+          'Invalid API key, please make sure to copy a valid API key from Plainly web-app and try again.',
+        );
+      }
+
+      let newApiKey = apiKey;
+      if (pin) {
+        const pinClass = new Pin(pin.first, pin.second, pin.third, pin.fourth);
+        const secret = pinClass.getPin();
```
ivansenic, line 67: why are we recreating pin here?

src/hooks/useSettingsReducer.ts
```diff
@@ -0,0 +1,90 @@
...
+      const newSettings = {
+        ...settings,
+        apiKey: {
+          key: newApiKey,
+          encrypted: !!pin,
+        },
+      };
+
+      await setSettings(newSettings);
+    },
```
ivansenic, line 85: is this correct, that we first dispatch and then update the settings on disk?

imo save should be done in the event handler? or?

src/components/settings/PinInput.tsx
```diff
@@ -0,0 +1,126 @@
...
+  const fourthInputRef = useRef<HTMLInputElement>(null);
+
+  const changeDigit = useCallback(
+    (
+      digit: 'first' | 'second' | 'third' | 'fourth',
+      value: string | undefined,
+    ) => {
+      const parsedValue = value ? Number.parseInt(value) : undefined;
+
+      onChange(
```
ivansenic, line 29: not sure what's going on but once I enter two digits quite fast, it does not move to another field..

ivansenic (changes requested):
Seems to be working fine.. Let's resolve other comments please..

We should crate a ticket for that plugin settings restore that can apparently come outside the app, from that zxp installer.. there's securely some event we would need to handle or something like that

src/components/settings/PinInput.tsx
vladavoX: it is because there was already a number entered, for example 1, and if you type 1 again in it, it doesn't trigger onChange because nothing was changed, using onInput instead to fix 👍

src/node/settings.ts
vladavoX: yes, not needed

src/node/settings.ts
ivansenic: OK if that works it's even better

src/node/settings.ts
ivansenic: yes, let's have in one single place not hardcoded in code

src/hooks/useSettingsReducer.ts
vladavoX: needed for the fetch settings, to get them from file and put them in state

src/hooks/useSettingsReducer.ts
vladavoX: as agreed, needed in useEffect

vladavoX:
regarding this, I will ask somewhere
- We should crate a ticket for that plugin settings restore that can apparently come outside the app, from that zxp installer.. there's securely some event we would need to handle or something like that

danixeee:
@vladavoX Can you add * or something when you type PIN?

plainly-plugin/src/node/constants.ts
```diff
@@ -0,0 +1,34 @@
...
+);
+const device: string = os.platform();
+function getSettingsDirectory(): string {
+  if (device === 'win32') {
+    return windowsDest;
+  }
+  return macDest;
+}
+
+export const settingsDirectory = getSettingsDirectory();
```
danixeee, line 34: `export const settingsDirectory = device === 'win32' ? windowsDest : macDest`, no need for a function

plainly-plugin/src/node/constants.ts
```diff
@@ -0,0 +1,34 @@
...
+export const csInterface = new CSInterface();
+
+// directory related
+const homeDirectory = os.homedir();
+const macDest = path.join(
+  homeDirectory,
+  'Library',
+  'Application Support',
+  'Adobe',
+  'Plainly Videos',
```
danixeee, line 17: I suggest using dir names without spaces when possible. So here and in `windowsDest` path as well

vladavoX:
> @vladavoX Can you add * or something when you type PIN?

Senic wanted without *

plainly-plugin/src/node/constants.ts
vladavoX: these are the locations where app data is stored in both win and mac, don't know where should we put it instead. Why are spaces a problem tho?

plainly-plugin/src/node/settings/settings.ts
```diff
@@ -0,0 +1,40 @@
...
+
+import { settingsDirectory } from '../constants';
+import type { Settings } from '../types';
+
+export const defaultSettings: Settings = {};
+
+async function retrieveSettings() {
+  try {
+    const file = await fsPromises.readFile(
+      path.join(settingsDirectory, 'settings.json'),
```
ivansenic, line 13: so twice `settings.json`, twice path created, wouldn't it be easier that settingsFileName and settingsPath are in the constants? and actually only path can be exported..

plainly-plugin/src/node/request.ts
```diff
@@ -0,0 +1,70 @@
...
+          try {
+            const response = JSON.parse(data);
+            const { statusCode } = res;
+            if (statusCode >= 200 && statusCode < 400) {
+              resolve(response);
+            } else {
+              reject({
+                statusCode,
+                message: data,
+              });
```
ivansenic, line 38: * so here rejects has one format, and message is a complete body object
* below line `41` rejects is done with string only
* and then on `48` it's again an object, with additional error (which is fine)

is this a good design? how the hack would you do anything with that reject?

plainly-plugin/src/ui/components/settings/SettingsForm.tsx
```diff
@@ -0,0 +1,236 @@
...
+        )}
+
+        {!settingsLoading && (
+          <div className="grid grid-cols-1 gap-x-6 gap-y-8 sm:grid-cols-6">
+            <div className="col-span-full">
+              <div>
+                <Label label="API key" htmlFor="api-key" required />
+                <Description>
+                  Setup your Plainly Videos API key to communicate directly with
+                  Plainly app. Your API key can be found in the{' '}
```
ivansenic, line 108:
```suggestion
                  Plainly app. Your API key can be found on the{' '}
```

plainly-plugin/src/ui/hooks/useSettings.ts
```diff
@@ -0,0 +1,95 @@
...
+      const newSettings = {
+        ...settings,
+        apiKey: {
+          key: newApiKey,
+          encrypted: !!pin,
+        },
+      };
+
+      try {
+        await saveSettings(newSettings);
```
ivansenic, line 82: * as said why is this not done in the `settingsReducer`
* try catch could be moved there as well
* tomorrow when we have new action you don't have to do anything

```
function settingsReducer(settings: Settings, action: Action) {
  const update  = switch (action.type) {
    case 'SET_SETTINGS': {
      return action.payload;
    }
    case 'SET_SETTINGS_API_KEY': {
      return {
        ...settings,
        apiKey: action.payload,
      };
    }
    default:
      return settings;
  }

  // save to disk here if not `SET_SETTINGS' (INIT_SETTINGS)`, right?

  return update;
}
```

plainly-plugin/src/ui/hooks/useSettings.ts
```diff
@@ -0,0 +1,95 @@
...
+  defaultSettings,
+  retrieveSettings,
+  saveSettings,
+} from '../../node/settings';
+import type { Settings } from '../../node/types';
+import type { Pin } from '../types';
+
+type Action =
+  | {
+      type: 'SET_SETTINGS';
```
ivansenic, line 14: this should be `INIT_SETTINGS`

ivansenic (approved):
Let's get this done and move forward..

plainly-plugin/src/ui/hooks/useSettings.ts
vladavoX: that would require me to use `await` there -> settingsReducer needs to become async -> useReducer dosen't accpet async functions.

plainly-plugin/src/node/request.ts
vladavoX: okay, so `mdn` docs say that we should make a `reject(reason: Error) => void`. So after that I had to make some changes, take a look

plainly-plugin/src/node/request.ts
```diff
@@ -0,0 +1,72 @@
...
+        res.on('end', () => {
+          try {
+            const response = JSON.parse(data);
+            const { statusCode, statusMessage } = res;
+            if (statusCode >= 200 && statusCode < 400) {
+              resolve(response);
+            } else {
+              reject({
+                name: 'ResponseError',
+                message: statusMessage,
```
ivansenic, line 37: what's status message? I would go for:

```
response?.message || statusMessage
```

ivansenic (approved):

plainly-plugin/src/node/request.ts
ivansenic: yes it's good now

PR #19, opened 2024-12-23
closes #18: Fix Export fails if folder already exists

### LINKED ISSUE #18: Export fails if the folder already exists
* folder where assets are added should have random name
* it will anyway be deleted at the end, and ZIP only contains the folder data and not the folder itself

image

vladavoX:
Original issue: #18

### DIFF
--- a/script/updateManifest.cjs
+++ b/script/updateManifest.cjs
@@ -37,8 +37,8 @@ fs.readFile(manifestPath, 'utf-8', (err, data) => {
     .replace(
       /<Menu>(.*?)<\/Menu>/,
       isDev
-        ? '<Menu>Plainly plugin - Dev mode</Menu>'
-        : '<Menu>Plainly plugin</Menu>',
+        ? '<Menu>Plainly Videos - Dev mode</Menu>'
+        : '<Menu>Plainly Videos</Menu>',
     );
 
   if (updatedManifest === data) {
--- a/src/components/form/ExportForm.tsx
+++ b/src/components/form/ExportForm.tsx
@@ -27,13 +27,13 @@ export default function ExportForm() {
     setLoading(true);
     if (targetPath) {
       try {
-        const [result, projectName] = await collectFiles(targetPath);
-        await zip(result, projectName);
-        await removeFolder(result);
+        const { collectFilesDir, projectName } = await collectFiles(targetPath);
+        await zip(collectFilesDir, projectName);
+        await removeFolder(collectFilesDir);
         setZipStatus({
           title: 'Successfully zipped',
           type: 'success',
-          description: `Zip file created at: ${decodeURI(result)}`,
+          description: `Zip file created at: ${decodeURI(collectFilesDir)}`,
         });
         setLoading(false);
       } catch (error) {
--- a/src/components/form/ExportForm.tsx
+++ b/src/components/form/ExportForm.tsx
@@ -33,7 +33,7 @@ export default function ExportForm() {
         setZipStatus({
           title: 'Successfully zipped',
           type: 'success',
-          description: `Zip file created at: ${decodeURI(collectFilesDir)}`,
+          description: `Zip file created at: ${decodeURI(targetPath)}`,
         });
         setLoading(false);
       } catch (error) {
--- a/src/components/form/ExportForm.tsx
+++ b/src/components/form/ExportForm.tsx
@@ -27,8 +27,8 @@ export default function ExportForm() {
     setLoading(true);
     if (targetPath) {
       try {
-        const result = await collectFiles(targetPath);
-        await zip(result);
+        const [result, projectName] = await collectFiles(targetPath);
+        await zip(result, projectName);
         await removeFolder(result);
         setZipStatus({
           title: 'Successfully zipped',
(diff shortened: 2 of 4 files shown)


src/node/export.ts
```diff
@@ -2,6 +2,7 @@ const fs = require('fs');
 const fsPromises = require('fs/promises');
 const archiver = require('archiver');
 const path = require('path');
+const crypto = require('crypto');
```
ivansenic, line 5: why was this added?

src/node/export.ts
```diff
@@ -119,11 +121,15 @@ async function copyFootage(footage: Footage[], projectDir: string) {
...
  * @param zipPath The path of the directory to zip.
  */
-function zip(zipPath: string): Promise<void> {
+function zip(zipPath: string, projectName: string): Promise<void> {
   return new Promise((resolve, reject) => {
     const pathResolved = finalizePath(zipPath); // Normalize and resolve
 
-    const outputZipPath = `${pathResolved}.zip`;
+    // replace final part of path with project name to create zip with project name
+    const pathReplaced = path.join(path.dirname(pathResolved), projectName);
```
ivansenic, line 129: what's the meaning of the `pathReplaced`? even the comment above is completely unclear..

src/node/export.ts
```diff
@@ -33,24 +34,25 @@ function selectFolder(callback: (result: string) => void) {
  * @param targetPath the path of the folder where the files will be copied.
  * @param callback a callback that will be called with the project name.
  */
-async function collectFiles(targetPath: string): Promise<string> {
+async function collectFiles(targetPath: string): Promise<[string, string]> {
```
ivansenic, line 37: Is this the best way of returning a result? What would be wrong with the type
```
{
   collectedFilesDir: string;
   projectName: string;
}
```

src/node/export.ts
```diff
@@ -119,11 +121,15 @@ async function copyFootage(footage: Footage[], projectDir: string) {
  *
  * @param zipPath The path of the directory to zip.
  */
-function zip(zipPath: string): Promise<void> {
+function zip(zipPath: string, projectName: string): Promise<void> {
```
ivansenic, line 124: * `zipPath` is also wrong, should be `targetPath` or something like that
* can you please go over naming here so other people would know what's going on when they are looking at the code

src/node/export.ts
```diff
@@ -33,24 +34,25 @@ function selectFolder(callback: (result: string) => void) {
...
   await fsPromises.copyFile(finalizePath(projectInfo.projectPath), dest);
 
   await copyFonts(projectInfo.fonts, projectDir);
   await copyFootage(projectInfo.footage, projectDir);
 
-  return projectDir;
+  return [projectDir, projectName];
 }
 
 async function copyFonts(fonts: Fonts[], projectDir: string) {
```
ivansenic, line 58: also `targetDir` here..

src/components/form/ExportForm.tsx
```diff
@@ -27,8 +27,8 @@ export default function ExportForm() {
     setLoading(true);
     if (targetPath) {
       try {
-        const result = await collectFiles(targetPath);
-        await zip(result);
+        const [result, projectName] = await collectFiles(targetPath);
```
ivansenic, line 30: I get a wrong notification

image

ivansenic (changes requested):

src/node/export.ts
vladavoX: so it generates random uuid, crypto is nodes module

src/node/export.ts
```diff
@@ -114,16 +115,20 @@ async function copyFootage(footage: Footage[], projectDir: string) {
...
   return new Promise((resolve, reject) => {
-    const pathResolved = finalizePath(zipPath); // Normalize and resolve
+    const targetPathResolved = finalizePath(targetPath); // Normalize and resolve
 
-    const outputZipPath = `${pathResolved}.zip`;
+    // replace final part of path (random uuid) with project name to create zip with projectName.zip
+    const zipPath = path.join(path.dirname(targetPathResolved), projectName);
+    const finalizeZipPath = finalizePath(zipPath); // Normalize and resolve
+
+    const outputZipPath = `${finalizeZipPath}.zip`;
```
ivansenic, line 131: imo extension should already be there, right? same like on line `41`

```
    // replace final part of path (random uuid) with project name to create zip with projectName.zip
    const zipPath = path.join(path.dirname(targetPathResolved), `${projectName}.zip`);
    const outputZipPath = finalizePath(zipPath); // Normalize and resolve
```

ivansenic (approved):
Can you please test once again before merging, it takes too much time for me to start the dockur..

vladavoX:
### When it exists

image
image

### when it doesn't, and also notification message
image

script/updateManifest.cjs
```diff
@@ -37,8 +37,8 @@ fs.readFile(manifestPath, 'utf-8', (err, data) => {
     .replace(
       /<Menu>(.*?)<\/Menu>/,
       isDev
-        ? '<Menu>Plainly plugin - Dev mode</Menu>'
-        : '<Menu>Plainly plugin</Menu>',
+        ? '<Menu>Plainly Videos - Dev mode</Menu>'
+        : '<Menu>Plainly Videos</Menu>',
```
vladavoX, line 41: @ivansenic from your PR, needs to be changed here also, so `yarn build/dev` produce the same name

Issue #8, opened 2024-12-12
Add feature to upload the project as new Plainly project


ivansenic:
* Needs setup for the API key as part of settings.ini
   * We should explore if the key can be somehow considered as secret and encrypted
* Create the upload route with the form that accepts only name, description and tags 
   * Execute zip action in the temporary folder
   * After zip is done, upload as the new project

PR #26, opened 2025-01-15
relates to #8: Project upload without encrypted key

vladavoX:
Original issue: #8 

image


plainly-plugin/src/node/request.ts
```diff
@@ -69,4 +91,19 @@ async function get(path: string, apiKey: string) {
   return request(getOptions);
 }
 
-export { get };
+async function post(path: string, apiKey: string, body: FormData | string) {
```
ivansenic, line 94: can we simply have two posts: `post` and `postFormData` and have no typeof then, it's also more readble in the code itself then

plainly-plugin/src/node/request.ts
```diff
@@ -1,22 +1,23 @@
...
-  headers: { 'Content-Type': 'application/json' },
 };
 
-function request(options: RequestOptions) {
+function request(options: RequestOptions, body?: FormData | string) {
   return new Promise((resolve, reject: (reason: Error) => void) => {
     const req: ClientRequest = protocol.request(
       options,
-      (res: ServerResponse) => {
+      (res: IncomingMessage) => {
```
ivansenic, line 20: so how did the types change now?

plainly-plugin/src/ui/components/about/About.tsx
```diff
@@ -28,10 +28,14 @@ export default function About() {
       label: 'Export zip',
       version: '1.0.0',
     },
+    {
+      label: 'Project upload',
+      version: '1.0.0',
```
ivansenic, line 33: how can this be :) retroactively adding feature nice

plainly-plugin/src/ui/components/form/UploadForm.tsx
```diff
@@ -0,0 +1,157 @@
...
+  const [loading, setLoading] = useState(false);
+  const { notification, notifySuccess, notifyError, clearNotification } =
+    useNotification();
+
+  const handleSubmit = async (e: React.FormEvent) => {
+    e.preventDefault();
+    setLoading(true);
+
+    try {
+      const { apiKey } = await retrieveSettings();
```
ivansenic, line 36: * why not useSettings, that's why we created that?
* useSettings can also provide maybe a function to get the apikey, no matter if the key is encryped or not (later on)

plainly-plugin/src/ui/components/form/UploadForm.tsx
```diff
@@ -0,0 +1,157 @@
...
+
+      const tempDest = path.join(settingsDirectory, 'temp');
+      if (!fs.existsSync(tempDest)) {
+        fs.mkdirSync(tempDest, { recursive: true });
+      }
```
ivansenic, line 41: imo this does not bellong here, let's create a `collectFiles()` without a directory that will create a temp one and call itself with argument?

plainly-plugin/src/ui/components/form/UploadForm.tsx
```diff
@@ -0,0 +1,157 @@
...
+
+      const { collectFilesDir, projectName } = await collectFiles(tempDest);
+      await zip(collectFilesDir, projectName);
```
ivansenic, line 44: let's have zip return the complete zip path, and not assume it on line `48`, also fix in other places where we call zip..

plainly-plugin/src/ui/components/form/UploadForm.tsx
```diff
@@ -0,0 +1,157 @@
...
+      if (tags) {
+        for (const tag of tags) {
+          formData.append('tags', tag);
+        }
+      }
+      formData.append('file', file);
+
+      await post('/api/v2/projects', apiKey.key, formData);
+
+      fs.rmSync(path.join(tempDest, `${projectName}.zip`), {
```
ivansenic, line 64: use path from zip

plainly-plugin/src/ui/components/form/UploadForm.tsx
```diff
@@ -0,0 +1,157 @@
...
+        recursive: true,
```
ivansenic, line 65: file can not be recursive

plainly-plugin/src/ui/components/form/UploadForm.tsx
```diff
@@ -0,0 +1,157 @@
...
+        force: true,
+      });
+
+      notifySuccess('Project uploaded');
+      setInputs({});
+      setLoading(false);
+    } catch (error) {
+      notifyError('Failed to upload project', (error as Error).message);
+      setLoading(false);
+    }
```
ivansenic, line 75:
```suggestion
    } catch (error) {
      notifyError('Failed to upload project', (error as Error).message);
    } finally {
      setLoading(false);
    }
```

plainly-plugin/src/ui/components/form/UploadForm.tsx
```diff
@@ -0,0 +1,157 @@
+import FormData from 'form-data';
```
ivansenic, line 1: does not work on first try:


image

ivansenic (changes requested):
Non function with name, without name, upload does not work..

plainly-plugin/src/node/request.ts
vladavoX: this is correct one, Ceda gave me wrong first time we were creating this

plainly-plugin/src/ui/components/form/UploadForm.tsx
vladavoX: without api key, should be good now

plainly-plugin/src/node/request.ts
danixeee, line 1: We should use axios. It didn't work for you because it used XMLHttpRequest and it was going through electron/browser (you can see that in network tab when debugging the plugin). When you change the "adapter" to "http" it will use nodejs http module. I haven't tested it for https in production, so try that out as well.

```typescript
import axios from "axios";

const instance = axios.create({
  adapter: "http",  // <------ this is the key
  baseURL: "http://localhost:8080",
});

async function post(path: string, apiKey: string, body: FormData | string) {
  await instance.post(path, body, {
    headers: {
      "Content-Type": "multipart/form-data",
    },
    auth: {
      username: apiKey,
      password: "",
    },
  });
}
```

now the request does not go through browsers that you can

ivansenic: Yes pleaseeee

plainly-plugin/src/node/collect/collect.ts
```diff
@@ -30,7 +30,18 @@ function selectFolder(callback: (result: string) => void) {
  *
  * @param targetPath the path of the folder where the files will be copied.
  */
-async function collectFiles(targetPath: string): Promise<CollectFilesResult> {
+async function collectFiles(
+  targetPath: string | undefined,
+): Promise<CollectFilesResult> {
+  if (!targetPath) {
+    const tempDest = path.join(settingsDirectory, 'temp');
```
danixeee, line 37: No need for recursion, you could just say `targetPath = tempDest`, but let's do it the other way.

Use system temp dir and uuid instead, like this:

```ts
import { tmpdir } from 'os';

const PLUGIN_TMP_DIR = join(tmpdir(), 'plainlyvideos');

async function collectFiles(...) { 
    const tempDest = path.join(PLUGIN_TMP_DIR, crypto.randomUUID()); 
}
```

benefits:
- if rmdir do not work for any reason, system will clear tmp dir after restarting pc
- uuid - to avoid a potential issue if rmdir fails

plainly-plugin/src/node/request.ts
```diff
@@ -1,72 +1,45 @@
...
-      reject({
-        name: 'RequestError',
-        message: error.message,
-      }); // Reject the Promise if there's an error
-    });
-
-    req.end();
+async function get(path: string, apiKey: string) {
+  return instance.get(path, {
+    headers: {
```
danixeee, line 13: You can put `content-type: app/json` header in `axios.create` as default, and override it where needed, e.g. in `postFormData`. Not important, but you will also add put with the app/json content-type in the next pr

plainly-plugin/src/node/collect/collect.ts
danixeee: Can you also make a function `makeTmpDir` and call it from the `UploadForm`? And here revert `targetPath` to be required. So, from the `UploadForm` component you will now what's happening without going into other functions

plainly-plugin/src/ui/components/form/ExportForm.tsx
```diff
@@ -24,25 +24,31 @@ export default function ExportForm() {
     e.preventDefault();
     setLoading(true);
     if (targetPath) {
+      let collectFilesDirValue: string | undefined;
+
       try {
         const { collectFilesDir, projectName } = await collectFiles(targetPath);
```
danixeee, line 30: Why we need `projectName` to be returned from `collectFiles`? Just do this `const projectName = path.basename(targetPath);` where needed.

plainly-plugin/src/ui/components/form/UploadForm.tsx
```diff
@@ -0,0 +1,170 @@
...
+  const disabled = loading || !apiKey?.key;
+
+  const handleSubmit = async (e: React.FormEvent) => {
+    e.preventDefault();
+    setLoading(true);
+    let collectFilesDirValue: string | undefined;
+    let zipPathValue: string | undefined;
+
+    try {
+      const { collectFilesDir, projectName } = await collectFiles(undefined);
```
danixeee, line 39: Same as for `ExportForm`

```ts
let tmpDir: string | undefined;
let collectFilesDir: string | undefined;
let zipPath: string | undefined;

try {
   ...
   tmpDir = await makeTempDir();
   collectFilesDir = await collectFiles(tmpDir);
   const projectName = path.basename(pathToAep);
   ...   
   zipPath = await zip(...)
   ...
finally {
   rmDirs...
}

```

plainly-plugin/src/node/collect/collect.ts
vladavoX: we already have uuid when creating a folder for export

plainly-plugin/src/ui/components/form/ExportForm.tsx
vladavoX: is that really betther tho?
because everytime we use `collectFiles` we need `projectName` to move forward, and now we are adding another import for path and assining another constant, while we can just as we do now, take if from collectFiles?

vladavoX: also, getting this
image

danixeee: But you also always `zip` after collecting files, but it does not mean `zip` should be inside the `collectFiles`, right?

I think that you should:
- change `collectFiles` to only collect files, maybe even call it `collectProject` in plugin code (and leave the name in aescripts code), because it returns path to the collected project dir
- add `makeProjectZip` function that will:
   - call collectProject
   - call getProjectName - or directly call path.basename
   - zip
   - return path to the zip
- then use `makeProjectZip` in both `Export` and `Upload` components and avoid (almost) duplicated code

plainly-plugin/src/ui/components/about/About.tsx
ivansenic: just add with `1.0.1`?

plainly-plugin/src/node/request.ts
```diff
@@ -1,72 +1,42 @@
...
-
-    req.on('error', (error) => {
-      reject({
-        name: 'RequestError',
-        message: error.message,
-      }); // Reject the Promise if there's an error
-    });
-
-    req.end();
+async function get(path: string, apiKey: string) {
```
ivansenic, line 14: we should allow the whole settings to be passed.. but never mind for now, once we need it we will add it..

plainly-plugin/src/node/collect/index.ts
```diff
@@ -1 +1,18 @@
+import path from 'path';
+
+import { evalScriptAsync } from '../utils';
+import { collectProjectFiles, zip } from './collect';
+
 export * from './collect';
+
+export const makeProjectZip = async (targetPath: string) => {
```
ivansenic, line 8: why is this separated in `index.ts`? just have it in collect, it's anyway that functionality?

plainly-plugin/src/ui/components/form/UploadForm.tsx
```diff
@@ -0,0 +1,170 @@
...
+  const disabled = loading || !apiKey?.key;
+
+  const handleSubmit = async (e: React.FormEvent) => {
+    e.preventDefault();
+    setLoading(true);
+    let collectFilesDirValue: string | undefined;
+    let zipPathValue: string | undefined;
+
+    try {
+      await makeTmpDir();
```
ivansenic, line 40: * not here, again why would this be here? create `makeProjectZip()` that creates a zip in tmp folder..
* make tmp dir removed
```
makeProjectZip() {
   await fsPromises.mkdir(TMP_DIR, { recursive: true });
   await makeProjectZip(TMP_DIR);
}
```

plainly-plugin/src/ui/hooks/useSettings.ts
```diff
@@ -86,5 +86,7 @@ export const useSettings = () => {
     [],
   );
 
-  return { settings, setSettingsApiKey, loading };
+  const getSettingsApiKey = () => settings.apiKey;
```
ivansenic, line 89: * add todo for decrypt

plainly-plugin/src/ui/components/form/UploadForm.tsx
```diff
@@ -0,0 +1,170 @@
...
+    projectName?: string;
+    description?: string;
+    tags?: string[];
+  }>({});
+  const [loading, setLoading] = useState(false);
+  const { notification, notifySuccess, notifyError, clearNotification } =
+    useNotification();
+  const { getSettingsApiKey, loading: settingsLoading } = useSettings();
+
+  const apiKey = getSettingsApiKey();
```
ivansenic, line 29: call in handle submit not here

plainly-plugin/src/ui/components/form/ExportForm.tsx
```diff
@@ -1,11 +1,11 @@
+import path from 'path';
```
ivansenic, line 1: does not work anymore:

image

plainly-plugin/src/ui/components/form/UploadForm.tsx
```diff
@@ -0,0 +1,170 @@
+import FormData from 'form-data';
```
ivansenic, line 1: again getting this:

image

ivansenic (changes requested):

plainly-plugin/src/ui/components/form/ExportForm.tsx
ivansenic: Setting api key also does not work anymore, I have a valid key

image

plainly-plugin/src/ui/components/form/UploadForm.tsx
vladavoX: need it here, so I can disable the button, if there is not apiKey

plainly-plugin/src/node/collectProject/collectProject.ts
```diff
@@ -169,4 +170,27 @@ async function removeFolder(targetPath: string) {
   }
 }
 
-export { collectFiles, removeFolder, selectFolder, zip };
+// Overload signatures
```
ivansenic, line 173: wtf

plainly-plugin/src/node/collect/index.ts
```diff
@@ -1 +0,0 @@
-export * from './collect';
```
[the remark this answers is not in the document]
ivansenic: just saw you also did not extend the about page, but we can do that when bumping the verision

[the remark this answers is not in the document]
vladavoX: renamed this, so we don't have `collect` here and in `aescripts`, easier to search for, suggested by @danixeee

PR #30, opened 2025-01-23
Fix fonts collection

vladavoX:
- Throw error for unsupported font extension
- Lowercase font extensions

### DIFF
--- a/plainly-plugin/src/node/collectProject/collectProject.ts
+++ b/plainly-plugin/src/node/collectProject/collectProject.ts
@@ -72,9 +72,20 @@ async function copyFonts(fonts: Fonts[], targetDir: string) {
 
     const dest = path.join(fontsDir, `${font.fontName}.${font.fontExtension}`);
     try {
-      return await fsPromises.copyFile(src, dest);
-    } catch (error) {
-      throw new Error(src);
+      // if the file doesn't end with .otf or .ttf, copy it, otherwise, throw an error
+      if (font.fontExtension === 'otf' || font.fontExtension === 'ttf') {
+        return await fsPromises.copyFile(src, dest);
+      }
+
+      throw new Error(
+        `Unsupported font format: ${font.fontExtension} for font ${font.fontName}`,
+      );
+    } catch (error: unknown) {
+      // Add `src` to the error message if it's not already present
+      if (!(error as Error).message.includes(src)) {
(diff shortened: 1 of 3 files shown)


plainly-plugin/src/node/collectProject/collectProject.ts
```diff
@@ -72,9 +72,20 @@ async function copyFonts(fonts: Fonts[], targetDir: string) {
...
+
+      throw new Error(
+        `Unsupported font format: ${font.fontExtension} for font ${font.fontName}`,
+      );
+    } catch (error: unknown) {
+      // Add `src` to the error message if it's not already present
+      if (!(error as Error).message.includes(src)) {
+        (error as Error).message += ` (Source Path: ${src})`;
+      }
+      throw error; // Re-throw the enriched error
```
ivansenic, line 88: if should create new error imo.. not sure what JS standards are

ivansenic (approved):
Seems fine, but I think @danixeee should also have a look

plainly-plugin/src/node/collectProject/collectProject.ts
```diff
@@ -72,9 +72,20 @@ async function copyFonts(fonts: Fonts[], targetDir: string) {
...
     try {
-      return await fsPromises.copyFile(src, dest);
-    } catch (error) {
-      throw new Error(src);
+      // if the file doesn't end with .otf or .ttf, copy it, otherwise, throw an error
+      if (font.fontExtension === 'otf' || font.fontExtension === 'ttf') {
+        return await fsPromises.copyFile(src, dest);
+      }
+
+      throw new Error(
```
danixeee, line 80: why you don't add src here?

plainly-plugin/src/node/collectProject/collectProject.ts
```diff
@@ -196,7 +207,7 @@ async function makeProjectZip(targetPath: string) {
     const zipPath = await zip(collectFilesDir, projectName);
     return { collectFilesDir, zipPath };
   } catch (error) {
-    throw new Error(`Failed to collect files: ${(error as Error).message}`);
+    throw new Error((error as Error).message);
```
danixeee, line 210: just remove try-catch, no need to throw new error with the same message

danixeee (approved):

Issue #24, opened 2025-01-10
Plugin improvements


vladavoX:
- <s>Add checkbox to open folder when zip is created [export]</s> ✅ 
- <s>Change notification to be /folder/projectName.zip [export]</s> ✅ 
- <s>save projectId and revision when uploading [upload]</s>
- <s>add option to remove api key [general]</s>
- Implement TanStack Query everywhere where it is needed [general]

vladavoX:
once we merge the PR, I will check if everything is done from here

PR #32, opened 2025-01-23
relates to #24: Add option to open folder on export

vladavoX:
Original issue: #24


ivansenic (changes requested):
Does not work on windows, nothing happens..

ivansenic (approved):
Pushed a win fix in https://github.com/plainly-videos/after-effects-plugin/pull/32/commits/27edfe4bdd6882f32b00b491376f32de20f2fe86, check again for mac

PR #33, opened 2025-01-29
Fix: stack multiple notifications

vladavoX:
image


plainly-plugin/src/ui/components/form/ExportForm.tsx
```diff
@@ -122,14 +122,10 @@ export default function ExportForm() {
...
 
-      {notification && (
-        <Notification
-          title={notification.title}
-          type={notification.type}
-          description={notification.description}
-          onClose={clearNotification}
-        />
-      )}
+      <NotificationsOverlay
```
ivansenic, line 125: why is this not added to `MainWrapper`?

plainly-plugin/src/ui/components/common/Notification.tsx
```diff
@@ -1,69 +1,107 @@
 import { Transition } from '@headlessui/react';
```
ivansenic, line 1: do we have auto-clear of info notifications now as on the UI?

ivansenic (commented):
Quick look

plainly-plugin/src/ui/components/common/Notification.tsx
vladavoX: - we don't have info notifications yet, as they are not used. Only success and error for now.
- I've added auto-clear to the success notifications, thinking that error should stay there, so you have time to read it, screenshot it, whatever. Maybe that one can auto-clear as well, but with higher timeout?

plainly-plugin/src/ui/components/form/ExportForm.tsx
vladavoX: will try

plainly-plugin/src/ui/components/form/ExportForm.tsx
```diff
@@ -122,14 +121,10 @@ export default function ExportForm() {
...
-          title={notification.title}
-          type={notification.type}
-          description={notification.description}
-          onClose={clearNotification}
-        />
-      )}
+      {/* <NotificationsOverlay
+        notifications={notifications}
+        onClose={clearNotification}
+      /> */}
```
ivansenic, line 127: I don't understand what this means now?

ivansenic (approved):
:rocket:

PR #41, opened 2025-02-14
Recreate destination folder if still selected but removed on machine by user

vladavoX:


plainly-plugin/src/node/collectProject/collectProject.ts
```diff
@@ -47,7 +47,9 @@ async function collectProjectFiles(targetPath: string): Promise<string> {
   const pathResolved = finalizePath(targetPath); // Normalize and resolve
 
   const folderName = crypto.randomUUID();
-  await fsPromises.mkdir(path.join(pathResolved, folderName));
+  await fsPromises.mkdir(path.join(pathResolved, folderName), {
```
ivansenic, line 50: do we call mkdir anywhere else? is so, let's do the same..

ivansenic (approved):

plainly-plugin/src/node/collectProject/collectProject.ts
vladavoX: I've checked already, but will check again just in case

PR #44, opened 2025-02-14
closes #37: Create useNavigate hook to handle navigation

### LINKED ISSUE #37: Create useNavigate hook
Original comment: https://github.com/plainly-videos/after-effects-plugin/pull/27#discussion_r1954287643

vladavoX:
- Add to MissingApiKey component before merge!


vladavoX:
Add to MissingApiKey component before merge!

vladavoX:
> Are you sure these are only two places where navigation is used?

yes, because only navigation was in sidebar, until we introtuced it in MissingAPiKey also

Issue #38, opened 2025-02-13
Improve api hooks


vladavoX:
add:
- error handling
- retries
- base url
- etc.

PR #49, opened 2025-02-18
closes #9: Add projects list page

### LINKED ISSUE #9: Mark local projects with Plainly reference
* Save information about the plainly project reference in the project file
    * Save information like: `id`, `revision`, `name`, and anything else that could be needed for correct syncing later on
* Create Projects route displaying all projects
    * Simple list with name, date, etc
    * Allow opening project in web browser
    * Allow marking the Plainly project reference as local project (show if already set)
* On project upload reference should be added on the new project creation immediately
    * Furthermore, if reference is marked, the upload should execute edit and not upload

vladavoX:


plainly-plugin/.env.development
vladavoX, line 1: eol

plainly-plugin/src/ui/components/settings/SettingsForm.tsx
```diff
@@ -4,7 +4,7 @@ import {
   useSettings,
 } from '@src/ui/hooks';
 import {
-  CheckCircleIcon,
+  CircleCheckIcon,
```
vladavoX, line 7: just to be consistent, this one is better, looking the same as X icon

plainly-plugin/src/ui/hooks/api/useProjectApi.ts
```diff
@@ -11,12 +11,37 @@ import { API_REFETCH_INTERVAL } from '.';
 
 const PROJECTS_CACHE_ROOT = 'projects';
 
+// refresh project list and search as long as one in analysis, each 15s
```
vladavoX, line 14: comment is wrong, not 15sec any more

plainly-plugin/src/ui/state/store.ts
vladavoX, line 1: revert this when done with PR

vladavoX (commented):
Need to do marking with local project

plainly-plugin/src/env.ts
```diff
@@ -5,3 +5,4 @@ export const isDev = process.env.PLUGIN_ENV === 'development';
 export const pluginBundleVersion = process.env.PLUGIN_BUNDLE_VERSION;
 
 export const baseURL = process.env.PLAINLY_API_BASE_URL;
```
ivansenic, line 7: nit `apiBaseUrl`?

plainly-plugin/src/ui/types/project.ts
```diff
@@ -10,6 +10,7 @@ export interface Project extends ModelBaseWithDates {
   };
   revisionHistory?: ProjectRevision[];
   analyzed: boolean;
+  size: number;
```
ivansenic, line 13: i don't feel size has any real relevance here.. I would like more to see if we have any templates, maybe creation or modification date.. Let's discuss

plainly-plugin/src/ui/utils.ts
```diff
@@ -20,3 +20,21 @@ export function handleLinkClick(link: string | undefined) {
   // @ts-expect-error
   cep.util.openURLInDefaultBrowser(link);
 }
+
+export function getSizeWithUnit(size: number): {
```
ivansenic, line 24: remove imo

plainly-plugin/src/ui/hooks/api/useProjectApi.ts
```diff
@@ -26,6 +51,24 @@ export const projectsCacheRemove = (client: QueryClient, projectId: string) => {
   client.removeQueries({ queryKey: [PROJECTS_CACHE_ROOT, projectId] });
 };
 
+export const useGetProjects = (apiKey: string) => {
+  const { isLoading, error, data } = useQuery({
+    queryKey: [PROJECTS_CACHE_ROOT],
+    queryFn: async () => {
+      const { data } = await get<Project[]>('/api/v2/projects', apiKey);
+      const sortedByLastModified = data.sort(
```
ivansenic, line 59: * imo sorting should be done in the component

plainly-plugin/src/ui/components/projects/ProjectsList.tsx
```diff
@@ -0,0 +1,41 @@
+import { useGetProjects } from '@src/ui/hooks';
```
ivansenic, line 1: Feedback:
* imo we should not use this table, but more simple single-row table
* Actions should be on the right as icons
   * mark as local, download, open in browser
   * these actions should have tooltips 
* I see no action to mark project as the local one?
   * This action should  show a confirmation modal if some project is already marked as local
* The currently project I am working on exists on platform, but I don't see it selected
   * I would put that one at the top of the list, somehow "selected" 
   * Of course button for marking it as local should be disabled
* From the data, I would keep only name and rendering state
   * Eventually for the selected project we could show more, like templates (but not sure)
   * Maybe from templates we can have links to online render form, again not sure
* Let's think how can auto-parametrization be incorporated here, is it this page or not.. What about render previews? A bit of brain storming needed imo

image

plainly-plugin/src/ui/routes/pages.ts
```diff
@@ -1,4 +1,10 @@
-import { FolderOutput, InfoIcon, SettingsIcon, UploadIcon } from 'lucide-react';
+import {
```
ivansenic, line 1: Did not know where to put this, so here:

* when collapsed menu, we need tooltips (`title`) on the links.. actually just add them always
* plainly logo should be way smaller, it's out of proportion comparing to rest of the elements.. plus if smaller, maybe the side bar expanded could end up being smaller as well, seems like a lot of space waisted

image

ivansenic (changes requested):
This needs more thinking and designing and not pure copy of the web platform..

plainly-plugin/src/ui/routes/pages.ts
vladavoX: okay, different PR this, because `title` doesn't work. So we need custom solution for tooltips

plainly-plugin/src/ui/components/projects/ProjectsList.tsx
vladavoX: so previews and autoparametrization, and all that detailed stuff -> Project details page

plainly-plugin/src/ui/components/projects/ProjectsList.tsx
```diff
@@ -0,0 +1,66 @@
...
+import { useProjectData } from '@src/ui/hooks/useProjectData';
```
ivansenic, line 2: * these should have more spacing and be buttons
* add the tooltip after rebase
* icon for the web should be as below
* is link icon the best, was thinking about star or pin, what do you think?

image
image

plainly-plugin/src/ui/components/projects/ProjectsList.tsx
```diff
@@ -0,0 +1,66 @@
...
+import { Routes } from '@src/ui/types';
```
ivansenic, line 3: * I think the selected one should be out of the list at the top somehow
* With clear selection, maybe some kind of the green ring applied
* Also more data with this one, for example I want to have local and remote revision, maybe templates could and then set of actions maybe as buttons in the bottom of the card
* yes should be some kind of card rather that a single row

Maybe we can discuss this quickly? :call_me_hand: 

image

ivansenic (commented):
Seems like we are getting there.. this is way better imo.. let's rebase, and deal with tooltip, plus the comments I gave.. did not check full code, but like 97%.. let's talk to agree maybe where to go and what to do next :+1:

plainly-plugin/src/ui/components/projects/ProjectsList.tsx
vladavoX: 4. well then star icon

plainly-plugin/src/ui/components/projects/LinkedProject.tsx
```diff
@@ -0,0 +1,120 @@
+import { useProjectData } from '@src/ui/hooks/useProjectData';
```
ivansenic, line 1: Really getting better.. We are almost there.. I will add set of feedback on this:

image

* The text and this headers are confusing.. My suggestions:
   * Top 
      * Custom Projects -> Plainly Project
      * Remove description, if we can n ot make anything better this is confusing
   * Linked project :+1:
      * `Working project is linked to the project on the Plainly platform.`  
   * Add In the same sense as the linked the Other projects sub-heading  above the list, also with some description
* Selected project
   * Icons are too big decrease
   * Maybe bold project title
   * I should have option to unlink or? If yes add issue or do if easy
   * Maybe templates above date now it's 3v1, then we can nicely add action main action buttons below
* List
  * I still have no tooltip on those icons

plainly-plugin/src/ui/components/projects/ProjectsListItem.tsx
```diff
@@ -0,0 +1,99 @@
...
+            <Label
+              label={project.name}
+              className="truncate w-24 xs:w-32 sm:w-48 md:w-64 ml-2"
+            />
+            <p className="whitespace-nowrap w-20 text-gray-400 ml-2">
+              {format(project.lastModified, 'PP')}
+            </p>
+          </div>
+          <div className="flex items-center gap-2 ml-2">
+            <ProjectAction icon={StarIcon} action={link} />
```
ivansenic, line 87: can we somehow  scroll back to the top of page once you link, cause when you have a huge list this project dissaperes and you are not scrolled back..

plainly-plugin/src/ui/components/projects/ProjectsList.tsx
```diff
@@ -0,0 +1,87 @@
...
+import { AuthContext } from '../settings/AuthProvider';
+import Description from '../typography/Description';
+import Label from '../typography/Label';
+import { LinkedProject } from './LinkedProject';
+import { ProjectsListItem } from './ProjectsListItem';
+
+export function ProjectsList() {
+  const { apiKey } = useContext(AuthContext);
+  const [projectData, setProjectData] = useProjectData();
+  const { isLoading, data } = useGetProjects(apiKey);
```
ivansenic, line 18: maybe a small improvement in the future that api hooks `useGetProjects` load api key themself?

plainly-plugin/src/ui/components/projects/ProjectAction.tsx
```diff
@@ -0,0 +1,35 @@
+import classNames from 'classnames';
+import type { LucideProps } from 'lucide-react';
+
+export function ProjectAction({
```
ivansenic, line 4: as mentioned add tooltip

ivansenic (approved):
Have some suggestions, but I will approve this.. Can have a quick look before merge.. Looks very nice now..

plainly-plugin/src/ui/components/projects/ProjectsList.tsx
vladavoX: yes, will add that here
https://github.com/plainly-videos/after-effects-plugin/issues/38

plainly-plugin/src/ui/components/projects/LinkedProject.tsx
vladavoX: I did this partialy
- Maybe templates above date now it's 3v1 ✅ , then we can nicely add action main action buttons below ❌ 

Imo, actions should be in top right, because they are small, and then longer project names can be visible more until truncated. If we put for example
`project.name ------- [icon] Templates 3`

then this project name will start to truncate much earlier

ivansenic (approved):
OK let's just fix icon confusion :0

PR #57, opened 2025-02-27
closes #56: Fix URI malformed error when creating a zip

### LINKED ISSUE #56: URI malformed error when exporting project with encoding characters in name
1. Create project with name `vladimir%aleksic.aep`
2. Click `Export`
3. This is what we get
Image
4. Zip function fails only, and in last finalize path `const outputZipPath = finalizePath(zipPath); // Normalize and resolve`

Now if we remove this final finalizing path, I think it is actually good, because we are then getting `vladimir%aleksic.zip`
Image

Also if we save the project as the `vladimir aleksic.aep`, everything is good
Image

vladavoX:


ivansenic (approved):
Tried few examples and it all works, let's ship it..

PR #58, opened 2025-02-27
Upgrade version 1.2.0

vladavoX:


plainly-plugin/src/ui/components/about/About.tsx
```diff
@@ -62,10 +66,10 @@ export function About() {
         </Description>
       </div>
       <div>
-        <Label label="Application features" />
+        <Label label="Plugin basics" />
```
ivansenic, line 69: basics? `Plugin features` seems way more clear

scripts/upgrade-version.sh
```diff
@@ -0,0 +1,42 @@
+#!/bin/bash
```
ivansenic, line 1: is this the correct directory for this, would assume this is either in `bin/` or root

ivansenic, line 1: file is not runnable `chmod +x upgrade-version.sh`

scripts/upgrade-version.sh
```diff
@@ -0,0 +1,42 @@
...
+
+# Check if the script is called with a version argument
```
ivansenic, line 3: does not work: 

```
after-effects-plugin/scripts$ ./upgrade-version.sh 1.2.1
sed: can't read ./package.json: No such file or directory
sed: can't read plainly-aescripts/package.json: No such file or directory
sed: can't read plainly-plugin/package.json: No such file or directory
sed: can't read plainly-plugin/.env.development: No such file or directory
sed: can't read plainly-plugin/.env.production: No such file or directory
sed: can't read plainly-plugin/.env.test: No such file or directory
sed: can't read ./CSXS/manifest.xml: No such file or directory
sed: can't read ./CSXS/manifest.xml: No such file or directory
Version updated to 1.2.1
```

ivansenic (changes requested):

scripts/upgrade-version.sh
vladavoX: even now when I moved it to root?

Because on platform if you are not in root dir when calling this, it doesn't work

ivansenic: I would rather do it correct, and then fix the platform as well.. should be a simple chatgpt prompt

upgrade-version.sh
```diff
@@ -0,0 +1,45 @@
+#!/bin/bash
```
ivansenic, line 1: seems extra white space added in the manifest:

image

ivansenic (approved):
Nice

upgrade-version.sh
vladavoX: script doesn't do that, must be from earlier

ivansenic: remove it then

PR #60, opened 2025-03-11
relates to #38: Improve API hooks

vladavoX:
- remove appmixer -> /me
- use apiKey directly in api hooks
- add base urls
- <s>[ ] add org name in Settings once /me is implement</s>

### DIFF
--- a/plainly-plugin/src/ui/hooks/api/useProfileApi.ts
+++ b/plainly-plugin/src/ui/hooks/api/useProfileApi.ts
@@ -4,7 +4,10 @@ import { useMutation } from '@tanstack/react-query';
 export const useUserProfile = () => {
   const { isPending, error, mutateAsync } = useMutation({
     mutationFn: async (apiKey: string) => {
-      const { data } = await get('/user/me', apiKey);
+      const { data } = await get<{
+        organizationId: string;
+        organizationName: string;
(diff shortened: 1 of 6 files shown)


plainly-plugin/src/ui/hooks/api/useProfileApi.ts
```diff
@@ -1,13 +1,12 @@
...
+const BASE_PATH = '/user';
+
 export const useUserProfile = () => {
   const { isPending, error, mutateAsync } = useMutation({
     mutationFn: async (apiKey: string) => {
-      const { data } = await get(
-        '/api/v2/integrations/appmixer/user-profile',
-        apiKey,
-      );
+      const { data } = await get(`${BASE_PATH}/me`, apiKey);
```
vladavoX, line 9: wait for this one to be done on platform

ivansenic:
Moving to @danixeee for a review

vladavoX:
@danixeee we won't be doing that orgName stuff. Will take to many changes, because we only have`apiKey` in <AuthProvider />, and we can't just wrap that `Settings` route into <AuthProvider />

danixeee (approved):
Ok, if we agree to work on supporting multiple organisations in the plugin, we can then add these info 👍

PR #70, opened 2025-03-26
Add project link to renders; Small changes to linked project look

vladavoX:
### what was changed SS
image

### whole thing SS
image

### DIFF
--- a/plainly-plugin/src/ui/components/projects/LinkedProject.tsx
+++ b/plainly-plugin/src/ui/components/projects/LinkedProject.tsx
@@ -12,7 +12,7 @@ import {
   VideoIcon,
   XCircleIcon,
 } from 'lucide-react';
-import { useCallback, useMemo, useState } from 'react';
+import { useCallback, useState } from 'react';
 import { ProjectAction } from '.';
 import { ConfirmationDialog, Tooltip } from '../common';
 import { Label } from '../typography';
@@ -76,7 +76,7 @@ export function LinkedProject({
     [project.id, openProjectRenders],
   );
 
-  const templatesLength = useMemo(() => project.templates.length, [project]);
+  const templatesLength = project.templates.length;
 
   return (
     <>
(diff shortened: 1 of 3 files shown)


plainly-plugin/src/ui/components/projects/LinkedProject.tsx
```diff
@@ -68,6 +71,13 @@ export function LinkedProject({
     [project.id, openInWeb],
   );
 
+  const openRenders = useCallback(
+    () => openProjectRenders(project.id),
+    [project.id, openProjectRenders],
+  );
+
+  const templatesLength = useMemo(() => project.templates.length, [project]);
```
danixeee, line 79: please remove `useMemo`... the purpose of this hook is to memoize heavy calculations, this is unnecessary overhead, `project.templates.length` is instant

plainly-plugin/src/ui/components/projects/ProjectsListItem.tsx
```diff
@@ -18,11 +18,13 @@ export function ProjectsListItem({
   project,
   linkProject,
   openInWeb,
+  openProjectRenders,
   linkedExists,
 }: {
   project: Project;
   linkProject: (data: ProjectData) => void;
   openInWeb: (id: string) => void;
+  openProjectRenders: (id: string) => void;
```
ivansenic, line 27: why are those passes as props, can be in the item itself..

ivansenic (approved):
Just some generic comments here, rest seems fine:

* the order unlink, open in web, renders seems odd to me somehow, but fine I guess
* I think this page and future page needs some kind of a refresh option, maybe in the header on the right, just so I can refresh stuff

plainly-plugin/src/ui/components/projects/ProjectsListItem.tsx
vladavoX: because they are used in both linkedProject and ProjectListItem

vladavoX: so instead of creating two, I just pass 1 as prop to two components

ivansenic: ok fine

PR #78, opened 2025-05-14
closes #74: Fix project packaging use relinking

### LINKED ISSUE #74: Footage files are not linked correctly
Opening someone elses project, says that there are missing files.

- we need to perform linking and save of the project
- copy the project to zip destination
- then revert the files (perform linking) and save of the project

Issue I've found, with relink -> we are creating an `uuid` folder, that doesn't exist later on. So once relink is done, it is looking later for 
`uuid/(Footage)/image.jpg`, and it gives missing files again

danixeee:
**footage missing in the original project**

image

**issue with long paths on windows**

image

### DIFF
--- a/plainly-aescripts/src/relink.ts
+++ b/plainly-aescripts/src/relink.ts
@@ -25,7 +25,6 @@ function relinkFootage(relinkData: RelinkData): void {
     }
 
     if (isWin() && fullPath.length > 255) {
-      alert('long path detected'); // TODO: removeme
       fullPath = `\\\\?\\${fullPath}`;
     }
 
--- a/plainly-aescripts/src/relink.ts
+++ b/plainly-aescripts/src/relink.ts
@@ -18,12 +18,18 @@ function relinkFootage(relinkData: RelinkData): void {
     }
 
     const itemId = item.id.toString();
-    const replacementFilePath = relinkData[itemId];
+    let fullPath = relinkData[itemId];
 
-    if (!replacementFilePath) {
+    if (!fullPath) {
       continue;
     }
-    const replacementFile = new File(replacementFilePath);
+
+    if (isWin() && fullPath.length > 255) {
+      alert('long path detected'); // TODO: removeme
+      fullPath = `\\\\?\\${fullPath}`;
+    }
+
+    const replacementFile = new File(fullPath);
     if (replacementFile.exists) {
       item.replace(replacementFile);
     }
--- a/plainly-aescripts/src/relink.ts
+++ b/plainly-aescripts/src/relink.ts
@@ -1,0 +1,40 @@
+function relinkFootage(): void {
+  app.beginUndoGroup('Relink Footage');
+
+  for (let i = 1; i <= app.project.numItems; i++) {
+    const item = app.project.item(i);
+    if (!(item instanceof FootageItem)) {
+      continue;
+    }
+    if (!(item.mainSource instanceof FileSource)) {
+      continue;
+    }
+
+    const originalFile = item.mainSource.file;
+    if (!originalFile) {
+      continue;
+    }
+
+    if (app.project.file == null) {
+      return;
+    }
+
+    const projectDir = app.project.file.parent.fsName;
+    const aeFolderPath = getFolderPath(item.parentFolder).replace(
+      'Root',
+      '(Footage)',
+    );
+    const itemAbsPath = pathJoin(projectDir, aeFolderPath, item.name);
+    // check if file exists
+    item.replace(new File(itemAbsPath));
+  }
+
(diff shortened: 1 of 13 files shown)


plainly-aescripts/src/relink.ts
```diff
@@ -0,0 +1,40 @@
...
+    item.replace(new File(itemAbsPath));
+  }
+
+  app.endUndoGroup();
+
+  app.project.save();
+}
+
+function undoFootage() {
+  app.executeCommand(16);
```
danixeee, line 38: maybe we should have a code that explicitly goes through the footage again and revert footage by footage.. tbd

plainly-plugin/src/node/collectProject/collectProject.ts
```diff
@@ -26,188 +19,131 @@ function selectFolder(callback: (result: string) => void) {
...
+    ]);
+
+    return zipPath;
+  } finally {
+    for (let i = undoStack.length - 1; i >= 0; i--) {
+      try {
+        await undoStack[i]();
+      } catch (undoErr) {
+        // TODO: What to do here?
+        console.error('Undo failed:', undoErr);
```
danixeee, line 143: if something happens here, we need to be sure that we can revert the whole project back... maybe we should zip .aep file before changing it - just in case

plainly-plugin/src/node/collectProject/collectProject.ts
```diff
@@ -26,188 +19,131 @@ function selectFolder(callback: (result: string) => void) {
...
+    // TODO: Show a missing files
+    throw new Error('Some footage files are missing from the project.');
+  }
 
-    const dest = path.join(fontsDir, `${font.fontName}.${font.fontExtension}`);
-    // if the file doesn't end with .otf or .ttf, copy it, otherwise, throw an error
-    if (['otf', 'ttf', 'ttc'].includes(font.fontExtension)) {
-      return await fsPromises.copyFile(src, dest);
-    }
+  // Throw in case of footage file length > 255
```
danixeee, line 61: TODO: remove this code, I found a solution for long paths on Win

plainly-plugin/src/node/utils.ts
```diff
@@ -2,14 +2,44 @@ import child_process from 'child_process';
...
+    return false;
+  }
+}
+
+/**
+ * Renames a file or directory if it exists at the given path.
+ * @param src Source path to rename.
+ * @param dest Destination path to rename to.
+ */
+export async function renameIfExists(src: string, dest: string): Promise<void> {
```
vladavoX, line 34: nit: this should be just `rename(...)`

and `exists` should be checked when invoking this function. Why do we want to invoke this, just to get inside if that tells us to `return` if there is nothing.

So `if (exists(...)) rename(...)`

vladavoX (approved):
Seems to work.
I've had 3 saves tho, 2 are for relink - undo relink and third one is in `collectFiles()`

But since there is now a check before whether project is saved or not, can we remove `app.project.save` from `collectFiles()`? Since on bigger projects it can be expensive, and 2 is better than 3
```ts
  let aepFilePath = await evalScriptAsync('getProjectPath()');
  if (!aepFilePath) throw new Error('Project not opened or not saved');
  aepFilePath = finalizePath(aepFilePath);

  const aepFileDir = path.dirname(aepFilePath);
  const aepFileName = path.basename(aepFilePath, '.aep');

  // 1. Collect project data
  const result = await evalScriptAsync('collectFiles()');
```

plainly-plugin/src/node/collectProject/collectProject.ts
```diff
@@ -191,23 +42,150 @@ async function removeFolder(targetPath: string) {
...
+
+  // 1. Collect project data
+  const result = await evalScriptAsync('collectFiles()');
+  if (!result) throw new Error('Failed to collect files');
+  const projectInfo: ProjectInfo = JSON.parse(result);
+
+  const hasLongFootagePaths = projectInfo.footage.some((item) => {
+    const itemPath = item.itemFsPath;
+    return itemPath.length > 255;
+  });
```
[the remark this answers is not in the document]
danixeee: entire path maximum is 255

plainly-plugin/src/node/utils.ts
danixeee: I made it on purpose because I reuse it on 4 places and it's more readable in the main code now

danixeee:
```
cini mi se da je prosla verzija - kad ti nije sacuvan projekat, on ti odmah otvori gde da ga sacuvas
u ovoj verziji samo izbaci error - failed to export because your project is not saved
```

comment from Staki

PR #79, opened 2025-05-15
closes #77: Don't allow export/upload when footage item is missing

### LINKED ISSUE #77: Do not allow export/upload if footage is missing
I would allow to click on the button, but will show the message. Disabling buttons is often confusing to people, even if there's a clear message.

Image

vladavoX:


danixeee:
added in #78

PR #85, opened 2025-06-04
closes #39: Check for project switches

### LINKED ISSUE #39: Update view when switching project
switching project does not upload the view, let's add a ticket for that

_Originally posted by @ivansenic in https://github.com/plainly-videos/after-effects-plugin/pull/27#discussion_r1954526455_

vladavoX:
- <s>[x] Display warning modal to reload extension when project changes</s> ❌ 
- [x] Save in context, and refresh data
- [x] Add warning when doing Upload?

@ivansenic @danixeee thoughts?

### DIFF
--- a/plainly-aescripts/src/project.ts
+++ b/plainly-aescripts/src/project.ts
@@ -9,9 +9,13 @@ function setProjectData(id: string, revisionCount: number) {
     ExternalObject.AdobeXMPScript = new ExternalObject('lib:AdobeXMPScript');
   }
 
-  const mdata = new XMPMeta!(project.xmpPacket); //get the project's XMPmetadata
+  if (!XMPMeta) {
+    return 'Error: XMPMeta is not available';
+  }
+
+  const mdata = new XMPMeta(project.xmpPacket); //get the project's XMPmetadata
   // update the Label project metadata's value
-  const schemaNS = XMPMeta!.getNamespaceURI('xmp');
+  const schemaNS = XMPMeta.getNamespaceURI('xmp');
   try {
     mdata.setProperty(schemaNS, propNameId, id);
     mdata.setProperty(
@@ -27,6 +31,12 @@ function setProjectData(id: string, revisionCount: number) {
   project.save();
 }
 
+interface ProjectData {
+  documentId: string;
+  id: string;
+  revisionCount: number;
+}
+
 function getProjectData() {
   const project = app.project;
 
@@ -35,11 +45,15 @@ function getProjectData() {
     ExternalObject.AdobeXMPScript = new ExternalObject('lib:AdobeXMPScript');
   }
 
-  const mdata = new XMPMeta!(project.xmpPacket); //get the project's XMPmetadata
-  const schemaNS = XMPMeta!.getNamespaceURI('xmp');
-  const schemaNSMM = XMPMeta!.getNamespaceURI('xmpMM');
+  if (!XMPMeta) {
+    throw new Error('XMPMeta is not available');
+  }
 
-  const projectData = {
+  const mdata = new XMPMeta(project.xmpPacket); //get the project's XMPmetadata
+  const schemaNS = XMPMeta.getNamespaceURI('xmp');
+  const schemaNSMM = XMPMeta.getNamespaceURI('xmpMM');
+
+  const projectData: ProjectData = {
     documentId: String(mdata.getProperty(schemaNSMM, 'DocumentID')),
     id: String(mdata.getProperty(schemaNS, propNameId)),
     revisionCount: Number(mdata.getProperty(schemaNS, propNameRevisionCount)),
@@ -56,8 +70,12 @@ function removeProjectData() {
     ExternalObject.AdobeXMPScript = new ExternalObject('lib:AdobeXMPScript');
   }
 
-  const mdata = new XMPMeta!(project.xmpPacket); // get the project's XMPmetadata
-  const schemaNS = XMPMeta!.getNamespaceURI('xmp');
+  if (!XMPMeta) {
+    throw new Error('XMPMeta is not available');
+  }
+
+  const mdata = new XMPMeta(project.xmpPacket); // get the project's XMPmetadata
+  const schemaNS = XMPMeta.getNamespaceURI('xmp');
 
   mdata.deleteProperty(schemaNS, propNameId);
   mdata.deleteProperty(schemaNS, propNameRevisionCount);
--- a/plainly-aescripts/src/project.ts
+++ b/plainly-aescripts/src/project.ts
@@ -10,7 +10,7 @@ function setProjectData(id: string, revisionCount: number) {
   }
 
   if (!XMPMeta) {
-    return 'Error: XMPMeta is not available';
+    return 'Error: Project metadata functionality is unavailable';
   }
 
   const mdata = new XMPMeta(project.xmpPacket); //get the project's XMPmetadata
--- a/plainly-aescripts/src/project.ts
+++ b/plainly-aescripts/src/project.ts
@@ -76,3 +76,21 @@ function getProjectPath() {
 function saveProject() {
   app.project.save();
 }
+
+function getProjectBasicData() {
+  const project = app.project;
+
+  if (ExternalObject.AdobeXMPScript === undefined) {
+    // load the XMPlibrary as an ExtendScript ExternalObject
+    ExternalObject.AdobeXMPScript = new ExternalObject('lib:AdobeXMPScript');
+  }
+
+  const mdata = new XMPMeta!(project.xmpPacket); //get the project's XMPmetadata
+  const schemaNS = XMPMeta!.getNamespaceURI('xmpMM');
+
+  const data = {
+    documentId: String(mdata.getProperty(schemaNS, 'DocumentID')),
+  };
+
+  return JSON.stringify(data);
+}
--- a/plainly-aescripts/src/project.ts
+++ b/plainly-aescripts/src/project.ts
@@ -37,12 +37,10 @@ function getProjectData() {
 
   const mdata = new XMPMeta!(project.xmpPacket); //get the project's XMPmetadata
   const schemaNS = XMPMeta!.getNamespaceURI('xmp');
-
-  if (!mdata.doesPropertyExist(schemaNS, propNameId)) {
-    return undefined;
-  }
+  const schemaNSMM = XMPMeta!.getNamespaceURI('xmpMM');
 
   const projectData = {
+    documentId: String(mdata.getProperty(schemaNSMM, 'DocumentID')),
     id: String(mdata.getProperty(schemaNS, propNameId)),
     revisionCount: Number(mdata.getProperty(schemaNS, propNameRevisionCount)),
   };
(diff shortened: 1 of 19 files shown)


plainly-aescripts/src/project.ts
```diff
@@ -68,3 +68,21 @@ function removeProjectData() {
...
+
+function getProjectBasicData() {
+  const project = app.project;
+
+  if (ExternalObject.AdobeXMPScript === undefined) {
+    // load the XMPlibrary as an ExtendScript ExternalObject
+    ExternalObject.AdobeXMPScript = new ExternalObject('lib:AdobeXMPScript');
+  }
+
+  const mdata = new XMPMeta!(project.xmpPacket); //get the project's XMPmetadata
```
danixeee, line 80: can we somehow add a global type for `XMPMeta` to avoid `!` ?

danixeee, line 80: `project` can be undefined? I would check it here anyway

plainly-aescripts/src/project.ts
```diff
@@ -68,3 +68,21 @@ function removeProjectData() {
...
+  const schemaNS = XMPMeta!.getNamespaceURI('xmpMM');
+
+  const data = {
```
danixeee, line 83: make a type

```ts
type AeProjectMetadata = {
    documentId: string;
}
```

danixeee: but this should actually be "projectId", right?

`projectId: String(mdata.getProperty(schemaNS, 'DocumentID')),`

plainly-plugin/src/ui/App.tsx
```diff
@@ -30,6 +44,35 @@ export function App() {
...
   }, [currentPage]);
 
+  const handleDialogClose = useCallback(() => {
+    setProjectChanged(false);
+    setDismissedReload(true);
+  }, []);
+
+  useEffect(() => {
+    const interval = setInterval(async () => {
+      const data = await evalScriptAsync('getProjectBasicData()');
```
danixeee, line 54: put this code in a separate function that will call script and parse data

plainly-plugin/src/ui/App.tsx
```diff
@@ -58,7 +101,18 @@ export function App() {
         </p>
       </Banner>
 
-      <div className="flex-1 overflow-y-auto">{route}</div>
+      <div className="flex-1 overflow-y-auto">
```
danixeee, line 104: hmm, I don't get this. Why anyone should reload the extension if we can switch the context in the background?

danixeee (commented):
I think interval is not enough, we should also make sure that we have the right context when user clicks on export / upload

plainly-aescripts/src/project.ts
vladavoX: typescript says that `app.project` is Project, and can't be `undefined`. Do you still want me to add it or?

plainly-aescripts/src/project.ts
vladavoX: I think calling it as it is is better. Maybe we will need `projectId` for something that is really `project id` and we won't have it available

plainly-plugin/src/node/collectProject/collectProject.ts
```diff
@@ -143,11 +144,19 @@ async function makeProjectZip(targetPath: string): Promise<string> {
...
+        const footageExists = await exists(footageDir);
+        return footageExists ? removeFolder(footageDir) : undefined;
+      }),
+    );
 
     // 5. Copy project fonts to the Fonts folder
     await copyFonts(projectInfo.fonts, aepFileDir).finally(() =>
-      undoStack.push(() => removeFolder(fontsDir)),
+      undoStack.push(async () => {
+        const fontsExists = await exists(fontsDir);
```
danixeee, line 157: same as above

plainly-plugin/src/ui/components/form/UploadForm.tsx
```diff
@@ -89,9 +94,19 @@ export function UploadForm() {
 
       let project: Project | undefined = undefined;
 
+      // check if the `documentId` and project from `getData` have the same ID
+      const projectData = await getData();
+      if (documentId && projectData?.documentId !== documentId) {
```
danixeee, line 99: so again, what is documentId? Can you tell me how this process works? I thought we only care and store project id to the xml metadata

plainly-plugin/src/ui/components/form/UploadForm.tsx
```diff
@@ -89,9 +94,19 @@ export function UploadForm() {
...
+        notifyError(
```
danixeee, line 100: This should actually never happen right?

plainly-plugin/src/ui/components/projects/LinkedProject.tsx
```diff
@@ -28,7 +28,7 @@ export function LinkedProject({
   openInWeb: (id: string) => void;
   openProjectRenders: (id: string) => void;
 }) {
-  const [projectData] = useProjectData();
+  const plainlyProject = useContext(GlobalContext)?.plainlyProject;
```
danixeee, line 31: I think this is a bit more readable

```ts
const { plainlyProject } = useContext(GlobalContext) || {};
```

plainly-plugin/src/ui/components/projects/LinkedProject.tsx
```diff
@@ -108,7 +108,7 @@ export function LinkedProject({
                 <Tooltip text="Sync status">
                   <FolderSync className="size-3" />
                 </Tooltip>
-                <p>Local v{projectData?.revisionCount}</p>
+                <p>Local v{plainlyProject?.revisionCount}</p>
```
danixeee, line 111: if this is undefined we will see "Local vundefined". Should we use `0` as default or render <p> conditionally

plainly-aescripts/src/project.ts
```diff
@@ -35,14 +45,16 @@ function getProjectData() {
...
+  if (!XMPMeta) {
+    throw new Error('XMPMeta is not available');
   }
 
-  const projectData = {
+  const mdata = new XMPMeta(project.xmpPacket); //get the project's XMPmetadata
+  const schemaNS = XMPMeta.getNamespaceURI('xmp');
+  const schemaNSMM = XMPMeta.getNamespaceURI('xmpMM');
+
+  const projectData: ProjectData = {
```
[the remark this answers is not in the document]
vladavoX: so this `documentId` is chaning each time we change project in AE, while we may, or may not have other data, depending if project is uploaded or not

plainly-plugin/src/ui/components/form/UploadForm.tsx
vladavoX: https://github.com/plainly-videos/after-effects-plugin/pull/85#discussion_r2175054232 check this.

So, we don't need to have `projectId` and `revisionCount` stored, because they are only stored once we execute the upload. 

For example:
- Execute upload
- we store this data (old stuff)
- Change project
- UI has no clue that we've changed the project
- We "upload existing", while it is not existing, since we have no way to check inside of AE if project has changed (this does change on navigation or close/open plugin)
- We overwrite, instead of upload a new project.


Now, we've added an additional data, to check for actual "project id" (AE project ID, a.k.a `documentId`), so it would go like this

- Execute upload
- we store this data (old stuff) + `documentId` in the context
- Change project
- UI now has a clue that something has changed
- Then we fetch data (if it exists), for the new project (then we can also upload existing, but for a new one, not overwrite the one before), or just leave it blank (then user can Upload new, if data doesn't exist)
- We upload new, or upload existing (but actuall project)

plainly-plugin/src/ui/components/form/UploadForm.tsx
vladavoX: there is a delay of 3s, so yes, there is a chance, if user is really quick, but then we have this backup on Upload, that will catch him

plainly-aescripts/src/project.ts
```diff
@@ -9,9 +9,13 @@ function setProjectData(id: string, revisionCount: number) {
     ExternalObject.AdobeXMPScript = new ExternalObject('lib:AdobeXMPScript');
   }
 
-  const mdata = new XMPMeta!(project.xmpPacket); //get the project's XMPmetadata
+  if (!XMPMeta) {
+    return 'Error: Project metadata functionality is unavailable';
```
danixeee, line 13: Any reason why we add this "Error: " to the message?

I think we should `throw new Error(message)`. 

What functionality won't work if `XMPMeta` is not available? I think we should state that, example:

`throw new Error("Unable to link the project due to missing metadata")`

plainly-aescripts/src/project.ts
```diff
@@ -35,14 +45,16 @@ function getProjectData() {
     ExternalObject.AdobeXMPScript = new ExternalObject('lib:AdobeXMPScript');
   }
 
-  const mdata = new XMPMeta!(project.xmpPacket); //get the project's XMPmetadata
-  const schemaNS = XMPMeta!.getNamespaceURI('xmp');
-
-  if (!mdata.doesPropertyExist(schemaNS, propNameId)) {
-    return undefined;
+  if (!XMPMeta) {
+    throw new Error('XMPMeta is not available');
```
danixeee, line 49: the same message as above, we should say something meaningful to the end user

plainly-aescripts/src/project.ts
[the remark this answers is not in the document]
danixeee: ok

plainly-aescripts/src/project.ts
```diff
@@ -58,8 +70,12 @@ function removeProjectData() {
     ExternalObject.AdobeXMPScript = new ExternalObject('lib:AdobeXMPScript');
   }
 
-  const mdata = new XMPMeta!(project.xmpPacket); //get the project's XMPmetadata
-  const schemaNS = XMPMeta!.getNamespaceURI('xmp');
+  if (!XMPMeta) {
+    throw new Error('XMPMeta is not available');
```
danixeee, line 74: same

plainly-plugin/src/ui/components/common/ConfirmationDialog.tsx
```diff
@@ -79,7 +81,7 @@ export function ConfirmationDialog({
               </Button>
               <Button
                 type="button"
-                onClick={() => setOpen(false)}
+                onClick={() => (onClose ? onClose() : setOpen(false))}
```
danixeee, line 84: shouldn't we call both?

```ts
() => {
    onClose?.();
    setOpen(false);
}
```

plainly-plugin/src/ui/components/projects/ProjectsList.tsx
```diff
@@ -1,33 +1,35 @@
...
 
   const linkedProject = useMemo(
-    () => data?.find((p) => p.id === projectData?.id),
-    [data, projectData],
+    () => data?.find((p) => p.id === plainlyProject?.id),
+    [data, plainlyProject],
   );
 
-  const filteredData = useMemo(
+  const projectsWithoutLinkedProject = useMemo(
```
danixeee, line 23: `const unlinkedProjects` ?

danixeee (commented):
I tested and it works good, I have a few minor comments

danixeee (approved):
Seems to be working fine 👍

PR #91, opened 2025-07-30
Fix windows destinations

vladavoX:


plainly-aescripts/src/project.ts
```diff
@@ -86,7 +86,7 @@ function removeProjectData() {
 }
 
 function getProjectPath() {
-  return app.project.file?.absoluteURI;
+  return app.project.file?.fsName;
 }
```
vladavoX, line 90: once the one above was fixed, that this also didn't return a good path, same as above

plainly-plugin/src/node/utils.ts
vladavoX, line 1: ok, revert this, and test with it

plainly-plugin/src/ui/App.tsx
vladavoX, line 1: revert

plainly-plugin/src/ui/components/form/ExportForm.tsx
vladavoX, line 1: for looks, only decodeURI is enough imo, revert

plainly-aescripts/src/relink.ts
vladavoX, line 1: es3 doesn't support `endsWith` so relinking in 1.2.6 doesn't work

plainly-aescripts/src/relink.ts
```diff
@@ -18,9 +18,10 @@ function relinkFootage(relinkData: RelinkData): void {
     }
 
     // Important: Skip PSD (Photoshop) and AI (illustrator) files
+    const name = originalFile.fsName.toLowerCase();
     if (
-      originalFile.fsName.endsWith('.psd') ||
-      originalFile.fsName.endsWith('.ai')
+      name.substring(name.length - 4) === '.psd' ||
+      name.substring(name.length - 3) === '.ai'
```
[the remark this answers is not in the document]
vladavoX: this won't work

PR #96, opened 2025-09-09
All `yarn install`s `--immutable`

vladavoX:


package.json
```diff
@@ -3,9 +3,9 @@
   "version": "1.2.7",
   "author": "Plainly Videos",
   "scripts": {
-    "build": "cd plainly-aescripts && yarn install && yarn build && cd ../plainly-plugin && yarn install && yarn build",
-    "prepare-aescripts": "cd plainly-aescripts && yarn install && yarn build",
-    "prepare-plugin": "cd plainly-plugin && yarn install && yarn build",
+    "build": "cd plainly-aescripts && yarn install --immutable && yarn build && cd ../plainly-plugin && yarn install --immutable && yarn build",
+    "prepare-aescripts": "cd plainly-aescripts && yarn install --immutable && yarn build",
+    "prepare-plugin": "cd plainly-plugin && yarn install --immutable && yarn build",
```
ivansenic, line 8: this seems wrong, `build` is somehow doing stuff way off.. this should be adapted now.. imo:

build-all: prepare-aescripts prepare-plugin build (not sure if install is needed here before build then, but that also seems wrong)

then ci install build-all

ivansenic (changes requested):

Issue #81, opened 2025-05-21
Detect and show project issues


danixeee:
Detect and show project issues such as:
- project has text layers that have ALL CAPS option enabled
- project contains videos / images / PDF / whatever that is not currently supported
- think of other possible issues

ivansenic:
so this is still on going? @vladavoX

PR #98, opened 2025-09-23
relates to #81: Project validation initial setup with first `allCaps` check

vladavoX:
image

image

### DIFF
--- a/plainly-plugin/src/ui/components/validate/text/AllCaps.tsx
+++ b/plainly-plugin/src/ui/components/validate/text/AllCaps.tsx
@@ -1,10 +1,15 @@
 import { AeScriptsApi } from '@src/node/bridge';
+import { useProjectData } from '@src/ui/hooks';
 import {
   ProjectIssueType,
   type TextAllCapsEnabledIssue,
 } from '@src/ui/types/validation';
 import classNames from 'classnames';
-import { ChevronDownIcon, CircleQuestionMark } from 'lucide-react';
+import {
+  ChevronDownIcon,
+  CircleQuestionMark,
+  TriangleAlertIcon,
+} from 'lucide-react';
 import type React from 'react';
 import { Tooltip } from '../../common';
 
@@ -17,6 +22,8 @@ export function AllCaps({
   isOpen: ProjectIssueType | undefined;
   setIsOpen: React.Dispatch<React.SetStateAction<ProjectIssueType | undefined>>;
 }) {
+  const [, , , aeVersion] = useProjectData();
+
   const onLayerNameClick = async (layerId: string) => {
     await AeScriptsApi.unselectAllLayers();
     await AeScriptsApi.selectLayer(layerId);
@@ -38,13 +45,20 @@ export function AllCaps({
         <div className="flex items-center gap-2">
           <p>All Caps enabled</p>
           <Tooltip
-            text="Using ALL CAPS text may result in poor font rendering. Auto-fixable in After Effects >= 24.3."
+            text="Using ALL CAPS text may result in poor font rendering. Auto-fixable in After Effects ≥ 24.3."
             className="max-w-40 whitespace-break-spaces"
           >
             <div className="flex items-center justify-center cursor-help size-4 group">
               <CircleQuestionMark className="size-4 text-gray-400 group-hover:text-white duration-200" />
             </div>
           </Tooltip>
+          {aeVersion && aeVersion >= 24.3 ? null : (
+            <Tooltip text="Auto-fix for this issue is available in After Effects version 24.3 and later.">
+              <div className="flex items-center justify-center cursor-help size-4 group">
+                <TriangleAlertIcon className="size-4 text-gray-400 group-hover:text-white duration-200" />
+              </div>
+            </Tooltip>
+          )}
         </div>
         <div className="flex items-center gap-2">
           <div className="bg-white/10 flex items-center justify-center rounded-full size-4">
--- a/plainly-aescripts/src/index.ts
+++ b/plainly-aescripts/src/index.ts
@@ -1,5 +1,6 @@
 import { collectFiles, selectFolder } from './collect';
 import {
+  getAfterEffectsVersion,
   getProjectData,
   getProjectPath,
   removeProjectData,
@@ -19,6 +20,7 @@ const PlainlyAE = () => ({
   removeProjectData,
   getProjectPath,
   saveProject,
+  getAfterEffectsVersion,
   relinkFootage,
   validateProject,
   fixAllIssues,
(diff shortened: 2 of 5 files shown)


danixeee:
I would slightly change the UI:
- layer name should be layer path
- if we cannot fix it (might be possible in 2025 but I am not sure), at least we should allow clicking on layer name that has the issue and selecting in in the timeline panel (or project files panel)
- "allCaps" as issue name is not good enough, but there's no much space to write something more, so maybe a tooltip or some different design should be tested

Not working?
image

plainly-aescripts/src/validateProject.ts
```diff
@@ -0,0 +1,46 @@
+interface ProjectValidation {
+  textLayers?: {
+    allCaps: {
+      layerId: string;
+      layerName: string;
+    }[];
+  };
+}
+
+function checkTextLayers(): ProjectValidation['textLayers'] | undefined {
```
danixeee, line 10: `ProjectValidation['textLayers']` this is a bit weird, you can make something like this:

```ts
enum ProjectIssueType {
    AllCaps,
    SomethingElse
}

type ProjectIssue<T extends ProjectIssueType> = {
    type: T;
}

type ProjectLayerIssue<T extends ProjectIssueType> = ProjectIssue<T> & {
    layerPath: {id: string, name: string}[]; // so you can find it in the project
    layerName: string;
}


type TextAllCapsEnabledIssue = ProjectLayerIssue<ProjectIssueType.AllCaps>;

type AnyProjectIssue = TextAllCapsEnabledIssue;


// validate project
const projectIssues: AnyProjectIssue[] = checkProject(app.project);
```

plainly-aescripts/src/validateProject.ts
danixeee, line 1: I think we should make a better project structure:

- src/validation/index.ts <- you can put validateProject() here (import and call other validators)
- src/validation/types.ts
- src/validation/textValidators.ts
- ...

plainly-plugin/src/ui/components/common/Tooltip.tsx
danixeee, line 1: hmm, is this the way to fix it? what was the problem here?

plainly-plugin/src/ui/components/validate/text/AllCaps.tsx
```diff
@@ -0,0 +1,38 @@
...
+              <CircleQuestionMark className="size-3 cursor-help" />
+            </Tooltip>
+          </div>
+          <div className="col-span-2 px-1 py-1 w-full">
+            <button
+              type="button"
+              className="text-left underline truncate max-w-full"
+              onClick={() => onLayerNameClick(details.layerId)}
+            >
+              {details.layerName}
```
danixeee, line 31: my bad, let's go with the layer name only, because I have a long comp name:

Image

plainly-plugin/src/ui/components/context/GlobalProvider.tsx
```diff
@@ -61,6 +63,28 @@ export const GlobalProvider = ({ children }: { children: React.ReactNode }) => {
     return () => clearInterval(interval);
   }, [notifyInfo, projectData]);
 
+  useEffect(() => {
```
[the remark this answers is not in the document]
vladavoX: so, I opened it, and I didn't press anything on screen for few seconds, and I got the badge and issues there (the project you've sent me as a zip). Can you send me a vod of example, I might not be doing the right thing to reproduce

plainly-plugin/src/ui/components/common/Tooltip.tsx
danixeee: No, I think it's overkill to install a whole lib because of the tooltip only. it's fine if this is working now

plainly-plugin/src/ui/components/context/GlobalProvider.tsx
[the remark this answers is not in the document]
danixeee: Maybe it was collapsed or something, I will check again

plainly-plugin/src/ui/components/common/Tooltip.tsx
vladavoX: okay, we can introduce later if needed. They have shit ton of components, and are most popular lib atm, better than tailwind

plainly-plugin/src/ui/components/validate/Validations.tsx
```diff
@@ -0,0 +1,139 @@
...
+      flushSync(() => setLoading(true));
+      await nextFrame();
+
+      try {
+        const issues = await evalScriptAsync('validateProject()');
+        if (!issues) {
+          setProjectData((prev) => ({ ...prev, projectIssues: [] }));
+        } else {
+          const parsedIssues: AnyProjectIssue[] = JSON.parse(issues);
+          if (JSON.stringify(parsedIssues) !== JSON.stringify(projectIssues)) {
```
danixeee: lodash is equal?

plainly-aescripts/src/validation/textValidators.ts
```diff
@@ -0,0 +1,106 @@
+function checkTextLayers(): TextLayerIssues[] | undefined {
```
danixeee: no need for undefined here, just return empty list

plainly-aescripts/src/validation/textValidators.ts
```diff
@@ -0,0 +1,106 @@
...
+      const range = textDocument.characterRange(0, textDocument.text.length);
+      // this means that there are different attributes in the text per character,
+      // thus, we can check per-character basis to find if any character is all-caps
+      if (range.fontCapsOption === undefined) {
+        for (let k = 0; k < textDocument.text.length; k++) {
+          const cRange = textDocument.characterRange(k, k + 1);
+          if (cRange.isRangeValid && cRange.fontCapsOption === 11014) {
+            hasCharacterAllCaps = true;
+            textLayers.push({
+              type: 'AllCaps' as ProjectIssueType.AllCaps,
```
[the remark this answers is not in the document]
vladavoX: same answer as above, seems to not understand it...

plainly-aescripts/src/validation/textValidators.ts
```diff
@@ -0,0 +1,124 @@
...
+        continue;
+      }
+
+      const textDocument = layer.sourceText.value;
+      // check if characterRange exists in textDocument (After Effects 24.3+)
+      if (typeof textDocument.characterRange !== 'function') {
+        continue;
+      }
+
+      let hasCharacterAllCaps = false;
```
danixeee: I think you can also do this check

```
if (textDocument.allCaps) // this checks only the first char

```

for unsupported AE version and maybe to optimize a bit?

PR #105, opened 2025-12-05
Remove replace "Root" folder

danixeee:
Fix when AE is using different languages

### DIFF
--- a/plainly-plugin/src/node/collectProject/collectProject.ts
+++ b/plainly-plugin/src/node/collectProject/collectProject.ts
@@ -65,11 +65,7 @@ function makeNewRelinkData(
   return footage.reduce(
     (acc, item) => {
       const itemId = item.itemId.toString();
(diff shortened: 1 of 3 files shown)


plainly-plugin/src/node/collectProject/collectProject.ts
```diff
@@ -164,11 +160,15 @@ async function makeProjectZip(targetPath: string): Promise<string> {
...
+        makeNewRelinkData(projectInfo.footage, footageDir),
+      )})`,
     ).finally(() =>
       undoStack.unshift(async () => {
         evalScriptAsync(
-          `relinkFootage(${JSON.stringify(makeOriginalRelinkData(projectInfo.footage))})`,
+          `relinkFootage(${JSON.stringify(
+            makeOriginalRelinkData(projectInfo.footage),
+          )})`,
         );
```
danixeee: biome format

PR #110, opened 2025-12-22
Fix dev mode due to monorepo changes

vladavoX:


plainly-aescripts/.yarn/releases/yarn-3.8.7.cjs
vladavoX: leftover, not needed

danixeee (approved):

PR #117, opened 2026-01-12
closes #116: Don't allow project collection if fonts are missing

### LINKED ISSUE #116: Don't allow collecting of project if there are missing fonts
- Currently it does this, AE replaces missing fonts with default fonts
- This is okay, but we are still packing it as its wrong name, and then this creates Analysis error

Solution, don't allow it at all, and give error message

vladavoX:
look at the PostScript names and compare them, because `.font` is a PostScript name returned by AE, then we can look at the file where location is pointing to, read it, and compare two PostScript names, if they are the same, all good.

We can't do something like:
- file name comparison, beacuse you can change files
- we can't just confirm that file exists on fontLocation, because it is always pointing to default font AE uses to replace
- there isn't anything built in into AE, that would return us any info that font is replaced

### DIFF
--- a/plainly-plugin/src/node/collectProject/utils.ts
+++ b/plainly-plugin/src/node/collectProject/utils.ts
@@ -29,7 +29,7 @@ async function validateFonts(fonts: Font[]) {
 
   if (missingFonts.length > 0) {
     throw new Error(
-      `Fonts used in the project, are missing on the system:\n${missingFonts.join(', ')}. Please install them and try again. If the problem persists, try restarting After Effects.`,
+      `Fonts used in the project, are missing on the system:\n${missingFonts.join(', ')}. Please install them and try again. If the problem persists, try restarting After Effects after installation.`,
     );
   }
 }


plainly-plugin/src/node/collectProject/collectProject.ts
```diff
@@ -50,6 +58,40 @@ function validateFootage(footage: Footage[]) {
...
+
+  // Throw in case of fonts missing
+  const missingExtension = fonts.filter((item) => !item.fontExtension);
+  if (missingExtension.length > 0) {
+    const fonts = missingExtension.map((f) => f.fontName);
+    throw new Error(`Fonts are missing extensions:\n${fonts.join(', ')}`);
+  }
+
+  // Throw in case of fonts missing location
+  const missingLocation = fonts.filter((item) => !item.fontLocation);
```
vladavoX: this can't be `undefined`, but in description they say that it can be returned as empty string if not good, so I am adding a check for that

plainly-aescripts/src/index.ts
```diff
@@ -18,6 +23,9 @@ const PlainlyAE = () => ({
   getProjectPath,
   saveProject,
   relinkFootage,
+  getFontsByPostScriptName,
+  getFontsByFamilyNameAndStyleName,
```
vladavoX: I would still keep this exposed, and keep it done in `node` part as it is now, what do you think?

plainly-plugin/src/node/collectProject/utils.ts
```diff
@@ -0,0 +1,70 @@
...
+    .filter((item) => !item.fontExtension)
+    .map((font) => font.postScriptName);
+  if (missingExtensionNames.length > 0) {
+    throw new Error(
+      `Fonts are missing extensions:\n${missingExtensionNames.join(', ')}`,
+    );
+  }
+
+  // Throw in case of fonts missing location
+  const missingLocationNames = fonts
```
vladavoX: dedup

plainly-plugin/src/node/collectProject/utils.ts
```diff
@@ -0,0 +1,67 @@
+import type { Font, Footage } from 'plainly-types';
+import { AeScriptsApi } from '../bridge';
+
+async function validateFonts(fonts: Font[]) {
+  if (fonts.length === 0) return;
+
+  // Throw in case of fonts missing extension
+  const missingExtensionNames = fonts
```
danixeee: Any reason why we need to throw error in case of this?
I saw many times that some fonts have extension .otf, but are actaully .ttf, and they we installed without issues. You can try out to confirm

danixeee: if it works, than use any of these as default, e.g. "TTF"

plainly-aescripts/src/collect.ts
```diff
@@ -40,19 +40,16 @@ function collectFonts(): Font[] {
   for (let i = 0; i < comps.length; i++) {
     const layers = getTextLayersByComp(comps[i]);
     for (let j = 0; j < layers.length; j++) {
-      const fontName = layers[j].sourceText.value.font;
-      const fontExtension = new File(
```
vladavoX: makes sense to be in `node` part and handeled there using `path`

plainly-plugin/src/node/collectProject/copyFonts.ts
```diff
@@ -22,16 +22,11 @@ export async function copyFonts(fonts: Font[], targetDir: string) {
...
-      );
-      // if the file doesn't end with .otf or .ttf, copy it, otherwise, throw an error
-      if (['otf', 'ttf', 'ttc'].includes(font.fontExtension)) {
-        return await fsPromises.copyFile(src, dest);
-      }
+    const fontName = `${font.postScriptName}.${font.fontExtension}`;
+    const dest = path.join(newFontsDir, fontName);
+    // if the file doesn't end with .otf or .ttf, copy it, otherwise, throw an error
+    if (['otf', 'ttf', 'ttc'].includes(font.fontExtension)) {
+      return await fsPromises.copyFile(src, dest);
```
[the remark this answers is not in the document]
vladavoX: irrelevant

plainly-aescripts/src/collect.ts
```diff
@@ -36,23 +36,27 @@ function collectFiles(): string | undefined {
 function collectFonts(): Font[] {
   const comps = getAllComps(app.project);
   const fonts: Font[] = [];
+  const seen = {};
```
danixeee: can you make fonts map?

```
const fonts: { [psName: string]: Font } = {};
...
return Object.values(fonts);
```

that way you don't need "seen"

plainly-plugin/src/node/collectProject/copyFonts.ts
```diff
@@ -22,20 +22,16 @@ export async function copyFonts(fonts: Font[], targetDir: string) {
...
-        newFontsDir,
-        `${font.fontName}.${font.fontExtension}`,
-      );
-      // if the file doesn't end with .otf or .ttf, copy it, otherwise, throw an error
-      if (['otf', 'ttf', 'ttc'].includes(font.fontExtension)) {
-        return await fsPromises.copyFile(src, dest);
-      }
+    const ext = path.extname(src).toLowerCase();
+    const fontName = `${font.postScriptName}${ext}`;
+    const dest = path.join(newFontsDir, fontName);
```
danixeee: you can move this inside if statement

PR #120, opened 2026-01-21
Return error message from backend

vladavoX:
image

### DIFF
--- a/plainly-plugin/src/node/errors.ts
+++ b/plainly-plugin/src/node/errors.ts
@@ -59,7 +59,7 @@ export class PlainlyApiError extends Error {
     errors?: ApiErrorResponseData['errors'],
     name = 'PlainlyApiError',
   ) {
-    super(errorMessage || code.toString());
+    super(errorMessage || CODE_MESSAGE_MAP[code] || code.toString());
     this.code = code;
     this.generalError = generalError;
     this.errorMessage = errorMessage;
@@ -77,11 +77,7 @@ export class PlainlyApiError extends Error {
  */
 export class NoInternetConnectionApiError extends PlainlyApiError {
   constructor() {
-    super(
-      ErrorCode.GENERAL_NO_INTERNET_CONNECTION,
-      true,
-      'NoInternetConnectionApiError',
-    );
+    super(ErrorCode.GENERAL_NO_INTERNET_CONNECTION, true);
 
     // Set the prototype explicitly.
     // see https://github.com/Microsoft/TypeScript/wiki/FAQ#why-doesnt-extending-built-ins-like-error-array-and-map-work
@@ -94,7 +90,7 @@ export class NoInternetConnectionApiError extends PlainlyApiError {
  */
 export class NoAccessTokenApiError extends PlainlyApiError {
   constructor() {
-    super(ErrorCode.GENERAL_NO_ACCESS_TOKEN, true, 'NoAccessTokenApiError');
+    super(ErrorCode.GENERAL_NO_ACCESS_TOKEN, true);
 
     // Set the prototype explicitly.
     // see https://github.com/Microsoft/TypeScript/wiki/FAQ#why-doesnt-extending-built-ins-like-error-array-and-map-work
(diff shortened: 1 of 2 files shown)


plainly-plugin/src/node/errors.ts
```diff
@@ -17,3 +17,23 @@ export class CollectFontsError extends Error {
     this.errorPaths = errorPaths;
   }
 }
+
+export class PlainlyApiError extends Error {
+  status?: number;
+  data?: unknown;
```
ivansenic: you could type this, it's a known structure:

```
{
  "timestamp": "2026-01-22T08:13:01.282+00:00",
  "status": 404,
  "error": "Not Found",
  "message": "The project with the ID 01bc9858-d6a3-4a21-a3fc-c758c458db3 does not exist.",
  "path": "/api/v2/renders"
}
```

Just make everything optional imo

ivansenic: e.g. for validations:

```
{
  "timestamp": "2026-01-22T08:15:17.905+00:00",
  "status": 400,
  "error": "Bad Request",
  "message": "Validation failed for object='projectRenderDto'. Error count: 2",
  "errors": [
    {
      "codes": [
        "NotNull.projectRenderDto.projectId",
        "NotNull.projectId",
        "NotNull.java.lang.String",
        "NotNull"
      ],
      "arguments": [
        {
          "codes": [
            "projectRenderDto.projectId",
            "projectId"
          ],
          "arguments": null,
          "defaultMessage": "projectId",
          "code": "projectId"
        }
      ],
      "defaultMessage": "must not be null",
      "objectName": "projectRenderDto",
      "field": "projectId",
      "rejectedValue": null,
      "bindingFailure": false,
      "code": "NotNull"
    },
    {
      "codes": [
        "NotBlank.projectRenderDto.projectId",
        "NotBlank.projectId",
        "NotBlank.java.lang.String",
        "NotBlank"
      ],
      "arguments": [
        {
          "codes": [
            "projectRenderDto.projectId",
            "projectId"
          ],
          "arguments": null,
          "defaultMessage": "projectId",
          "code": "projectId"
        }
      ],
      "defaultMessage": "must not be blank",
      "objectName": "projectRenderDto",
      "field": "projectId",
      "rejectedValue": null,
      "bindingFailure": false,
      "code": "NotBlank"
    }
  ],
  "path": "/api/v2/renders"
}
```

plainly-plugin/src/node/request.ts
danixeee: @vladavoX @ivansenic  not sure if this is way to go, we already have a similar logic in the frontend code, can we use that?

danixeee (commented):
Ok I guess we agreed with this approach, but, for the end-user, this message on the screenshot is not meaningful - he can't change the way plugin calls our API.

We should implement a file logging and a way for the end-user to share this log with us if something is not working properly.

@ivansenic wdyt?

plainly-plugin/src/node/errors.ts
```diff
@@ -17,3 +17,243 @@ export class CollectFontsError extends Error {
     this.errorPaths = errorPaths;
   }
 }
+
+export enum ErrorCode {
```
ivansenic: why are you having this, wtf is going on.. remove all non-client related error codes you don't need them and we don't want to maintain this

plainly-plugin/src/ui/components/form/UploadForm.tsx
```diff
@@ -128,7 +128,7 @@ export function UploadForm() {
       setUploadMode('edit');
       setInputs({});
     } catch (error) {
-      notifyError('Failed to upload project', (error as Error).message);
+      notifyError('Failed to upload project', error);
```
ivansenic: so here we use `error` in notifyError and below we go for `error.message`?

plainly-plugin/src/ui/components/settings/AuthProvider.tsx
```diff
@@ -31,7 +31,8 @@ export const AuthProvider = ({ children }: { children: React.ReactNode }) => {
         getSettingsApiKey(pin);
         setPinStorage(pin);
       } catch (error) {
-        notifyError((error as Error).message);
+        const eMessage = error instanceof Error ? error.message : String(error);
```
ivansenic: here we extract message, why?

plainly-plugin/src/ui/hooks/useNotifications.ts
```diff
@@ -50,7 +51,8 @@ export const useNotifications = () => {
   );
 
   const notifyError = useCallback(
-    (title: string, description?: string) => {
+    (title: string, details?: unknown) => {
```
ivansenic: `unknown` really? should it be `Error | string`?

danixeee (approved):

PR #128, opened 2026-02-24
Fix timeout error

vladavoX:
- timeout aproach is not good, we should not block the user when his connections are interupted, or it takes too long to upload.
- check codes and then throw based on them

### DIFF
--- a/plainly-plugin/src/node/request.ts
+++ b/plainly-plugin/src/node/request.ts
@@ -13,6 +13,8 @@ import {
 } from './errors';
 
 const PLAINLY_ERROR_CODE_HEADER = 'X-PlainlyErrorCode'.toLowerCase();
+const REQUEST_TIMEOUT_MS = 30000; // 30 seconds
+const FORM_DATA_REQUEST_TIMEOUT_MS = 120000; // 2 minutes
 const NO_INTERNET_ERROR_CODES = new Set([
   'ENOTFOUND',
   'EAI_AGAIN',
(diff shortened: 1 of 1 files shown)


danixeee (commented):
## Code Review

### Overview

This PR removes the 5-second request timeout from the axios instance and replaces the simple `navigator.onLine` check with a more robust `isLikelyOfflineError()` function that detects various network error conditions.

### What's Good

- **Comprehensive error detection**: The `NO_INTERNET_ERROR_CODES` set covers common DNS/network failure codes
- **Defense in depth**: Multiple detection strategies (navigator.onLine, error codes, message parsing)
- **Clean separation**: The `isLikelyOfflineError()` helper is well-structured and easy to test

### Concerns

1. **Complete removal of timeout** - This is the main concern. Removing the timeout entirely means requests can hang indefinitely if:
   - The server accepts the connection but never responds
   - A corporate proxy holds the connection open
   - File uploads to a slow endpoint never complete
   
   **Suggestion**: Consider keeping a longer timeout (e.g., 30-60s for API calls) rather than removing it entirely, or use per-request timeouts for upload operations specifically.

2. **`navigator` availability in Node environment** - This code runs in `src/node/`. The `navigator` object is a browser API and won't exist in Node.js. While the code guards against this, it's effectively dead code here.

3. **Message parsing fragility** - String matching on error messages (`'network error'`, `'getaddrinfo enotfound'`, etc.) is fragile—messages can change between axios versions or be localized. Consider relying primarily on error codes.

4. **Missing error codes** - Consider adding:
   - `ETIMEDOUT` - connection timeout (especially relevant if you restore timeouts)
   - `ECONNREFUSED` - server not running
   - `ECONNRESET` - connection reset by peer

### Recommendations

1. Don't remove timeout entirely—use a longer default (30-60s) or remove only for upload operations
2. Add `ETIMEDOUT`, `ECONNREFUSED`, and `ECONNRESET` to `NO_INTERNET_ERROR_CODES`
3. Add unit tests for `isLikelyOfflineError()` with various error scenarios

PR #133, opened 2026-03-05
fix: Missing footage didn't report error

vladavoX:
image


plainly-plugin/src/node/collectProject/utils.ts
```diff
@@ -34,9 +34,8 @@ async function validateFonts(fonts: Font[]) {
   }
 }
 
-function validateFootage(footage: Footage[]) {
+function validateFootage(_: Footage[], missingFootage: MissingFootage[]) {
```
vladavoX: its okay to leave it as first unused prop, we will probably need it in future

PR #137, opened 2026-03-12
relates to #4647: allow scripts modifications via plugin

vladavoX:
- Solves 5 static scripts at the moment
- 1 dynamic script layer management
- single layer script modifications
- bulk scripts


plainly-plugin/src/ui/components/parameters/Parameters.tsx
```diff
@@ -0,0 +1,447 @@
...
+const scriptName = (type: ScriptType) => {
+  if (type === ScriptType.CROP) return 'Crop';
+  if (type === ScriptType.MEDIA_AUTO_SCALE) return 'Media auto scale';
+  if (type === ScriptType.TEXT_AUTO_SCALE) return 'Text auto scale';
+  if (type === ScriptType.SHIFT_IN) return 'Shift in';
+  if (type === ScriptType.SHIFT_OUT) return 'Shift out';
+  return type;
+};
+
+function SelectAllCheckbox({
```
vladavoX: this and below Parametrized layers should be in their own separate component

vladavoX (commented):
self review - round one

vladavoX (commented):
self round 2

danixeee (commented):
If project is not linked, we should show an information message saying that the project needs to be linked first. Currently, parameters starts and stop loading, but I am not sure what is loading?

When I upload the new project and go to this page, I don't know what to do, so we need a message for now saying that template should be created from the web app, and I would give a link to the template creation.

We need some refresh button on this page I guess? e.g. "Reload templates" or whatever

Responsive: Can we make "Add script" dropdown button or something and put all scripts there?
Image
But the whole "Bulk add new scripts" seems that taking a lot of space, any other ideas? Also, we will have "pre-defined composite script - crop + shift", so maybe we should group these somehow in dropdown button.

Script that have parameters are impossible to remove, because on click it opens to edit them, and you don't have X in the top right corner.


The biggest issue IMO:
1. Go to web app, create template with 1 parameter and 1 script
2. Go to plugin and load this template
3. Go to web app, add another script to parameter
4. Go to plugin, add different script to parameter
5. All changes made in step 3. are overwritten by 4.

Also, as you noticed before, we have this problem with caching on the frontend side. In any case, we need to implement versioning, so you know that you will not overwrite the latest changes if you are not looking at the latest version.


I haven't looked at the code yet, but I would fix above problems first.

plainly-plugin/src/ui/components/parametrization/Parametrization.tsx
```diff
@@ -0,0 +1,358 @@
...
+import { ScriptDialogs } from './ScriptDialogs';
+
+export function Parametrization() {
+  const { plainlyProject, contextReady } = useContext(GlobalContext) || {};
+  const { isLoading, data, refetch, isRefetching } = useGetProjectDetails(
+    plainlyProject?.id,
+  );
+  const { notifyError, notifySuccess } = useNotifications();
+
+  const [query, setQuery] = useState('');
```
vladavoX: rename to `templateQuery`

vladavoX (commented):
## Overview

This PR introduces the **Parametrization screen** — a new UI that lets users view and modify layer-level scripts (Crop, Shift In/Out, Media Auto Scale, Text Auto Scale, Layer Management) on templates linked to a Plainly project. It also adds a new `getLayerNamesByComp` AE bridge function used by the Shift script dialogs to populate layer name dropdowns.

---

## What's good

- **Type modeling is solid.** The `template.ts` type file uses discriminated unions well (`Layer`, `EditableScript`, `ScriptEditState`). The typed overloads for `getDefaultScript` catch mismatches at compile time.
- **Immutable state updates throughout.** All `setEditableLayers` calls use functional updates and spread correctly — no accidental mutations.
- **`Checkbox` refactor is clean.** Fixing hardcoded `id="comments"` and making `disabled` / `defaultChecked` configurable was the right call.
- **`reorderScripts` utility is pure.** No side effects, returns new array — easy to test in isolation.
- **`SelectAllCheckbox` indeterminate state** is handled correctly via `useEffect` + ref — this is the right pattern since `indeterminate` is a DOM property, not an HTML attribute.
- **DnD using `scriptType` as the drag ID** is valid since the UI prevents adding duplicate script types to the same layer.

---

## Issues

### 1. Dead code — disabled Edit button with TODO

`ParametrizedLayers.tsx` ~line 1578–1585:
```tsx
<button
  onClick={() => {}}
  ...
  disabled={true}
>
  <EditIcon className="size-3" />
</button>
```
This renders a permanently-disabled button with a `TODO` comment. Either remove it and track the feature in a ticket, or hide it entirely — shipping disabled UI without a visible explanation can confuse users.

### 2. `projectsCacheRemove` on template edit failure is too aggressive

In `useTemplateApi.ts`:
```ts
onError: (_: unknown, { projectId }: { projectId: string }) =>
  projectsCacheRemove(queryClient, projectId),
```
This evicts the entire project from the React Query cache on any save failure. If the user edits two templates back-to-back and the second save fails, they lose the first template's cached data. Consider invalidating the specific project query instead (e.g. `queryClient.invalidateQueries([PROJECTS_CACHE_ROOT, projectId])`), so the data is refreshed on next access without being wiped.

### 3. `posNotZero` uses AND — position with only one axis set is silently dropped

In `AutoScaleMediaScriptDialog.tsx`:
```ts
const posNotZero = position && position.posX !== 0 && position.posY !== 0;
```
If a user sets `posX = 100` and leaves `posY = 0`, `posNotZero` is `false` and the position is cleared on save. The placeholder says `0` which implies zero is a valid anchor. Consider using `||` (either axis non-zero) or keeping the position if either field has been explicitly touched.

### 4. `biome.json` line width override is a code smell

The `lineWidth: 120` override targets a single file (`parametrization/utils.ts`). Long lines there are caused by the overloaded function signatures for `getDefaultScript`. Rather than bending the formatter, consider collapsing the overloads into a single signature or splitting them differently. This avoids an ever-growing list of file-specific formatter exceptions.

### 5. `compId` only reads `compositions[0]`

In `ScriptDialogs.tsx`:
```ts
const compId = activeLayer?.compositions[0]?.id;
```
If a layer belongs to multiple compositions, the Shift dialog always uses the first one to populate the layer name list. This may be intentional given the current data, but worth a comment documenting the assumption since the type allows `compositions: { id: number; name: string }[]` (plural).

### 6. Duplicate `TEXT_AUTO_SCALE` add-guard in two places

The duplicate-add guard for `TEXT_AUTO_SCALE` exists both in `Parametrization.tsx` (`handleBulkScriptSelect`) and in `ParametrizedLayers.tsx` (`handleScriptSelect`). Both are correct — just noting the logic is duplicated. A shared helper would reduce risk of divergence.

---

## Minor notes

- `@tailwindcss/typography` removal: confirm it's unused in all packages before merging.
- `2xs` font size (`0.65rem`) added to Tailwind config — fine, worth keeping in sync with design tokens.
- The `put` helper generic `B` for body type aligns with Axios conventions — no issue.

plainly-aescripts/src/utils.ts
```diff
@@ -69,6 +69,23 @@ function getFolderPath(folder: FolderItem): string {
   return pathJoin(parentPath, folder.name);
 }
 
+/**
+ * @function getLayerNamesByComp
+ * @description Get all layer names in a given After Effects comp by ID
+ * @param {string} compId - The ID of the comp
+ * @returns {string} JSON array of layer names
+ */
+function getLayerNamesByComp(compId: string): string {
```
danixeee: `getCompLayers` that returns a list of simple layer objects (e.g. `[{"name": "layer1"}, ...]` or just rename to `getCompLayerNames`

plainly-plugin/src/ui/components/parametrization/LayerManagementScriptDialog.tsx
```diff
@@ -0,0 +1,62 @@
+import type { LayerManagementScript } from '@src/ui/types/template';
+import { ScriptType } from '@src/ui/types/template';
+import { LayersIcon } from 'lucide-react';
+import { useState } from 'react';
+import { Description, Label } from '../typography';
+import { ScriptDialogShell } from './ScriptDialogShell';
+
+const PARAMETER_NAME_REGEX = /^[^.]+$/;
```
danixeee: maybe extract to some `constants.ts` or somewhere else?

plainly-plugin/src/ui/components/parametrization/ParametrizedLayers.tsx
```diff
@@ -0,0 +1,481 @@
...
+  COMPOSITION_REPLACE_SOURCE: 'Composition replace source',
+};
+
+const scriptName = (type: ScriptType) => {
+  const raw = type as unknown as string;
+  return (
+    SCRIPT_NAMES[raw] ??
+    raw.charAt(0).toUpperCase() + raw.slice(1).toLowerCase().replace(/_/g, ' ')
+  );
+};
```
danixeee: what is this magic here? `type as unknown as string;`, can we fix this? is this because we do not support all scripts ATM?

plainly-plugin/src/ui/components/parametrization/ParametrizedLayers.tsx
```diff
@@ -0,0 +1,481 @@
...
+  if (layerType === 'DATA_EFFECT') return <SparklesIcon className={cs} />;
+  if (layerType === 'MEDIA') {
+    if (mediaType === 'image') return <ImageIcon className={cs} />;
+    if (mediaType === 'video') return <VideoIcon className={cs} />;
+    if (mediaType === 'audio') return <AudioLinesIcon className={cs} />;
+  }
+  return null;
+}
+
+const EDITABLE_SCRIPT_TYPES = new Set([
```
danixeee: I think you can merge into one Set. You can add all scripts to the `ScriptType`, and do something like this:

```ts
const SCRIPTS_CONFIG = {
    [ScriptType.CROP]: { "label": "Crop", "isSupported": true},
    ...
}
```

and then no need for `const scriptName = (type: ScriptType)` and similar stuff, put everything in one map

plainly-plugin/src/ui/components/parametrization/ScriptDialogs.tsx
```diff
@@ -0,0 +1,184 @@
...
+} from '@src/ui/types/template';
+import { useCallback, useMemo } from 'react';
+
+import { AutoScaleMediaScriptDialog } from './AutoScaleMediaScriptDialog';
+import { CropScriptDialog } from './CropScriptDialog';
+import { LayerManagementScriptDialog } from './LayerManagementScriptDialog';
+import { ShiftScriptDialog } from './ShiftScriptDialog';
+import { getDefaultScript } from './utils';
+
+const SCRIPT_LAYER_TYPE_RESTRICTIONS: Partial<Record<ScriptType, LayerType[]>> =
```
danixeee: I would also put this into the same map: https://github.com/plainly-videos/after-effects-plugin/pull/137/changes#r3062820945

danixeee: and maybe move it to some common file for the scripts, "scriptRegistry" or similar

plainly-plugin/src/ui/components/parametrization/ScriptsDialog.tsx
```diff
@@ -0,0 +1,163 @@
...
+  LayersIcon,
+  LogInIcon,
+  LogOutIcon,
+  ScissorsIcon,
+  TypeIcon,
+} from 'lucide-react';
+import { Button } from '../common';
+import { Description } from '../typography';
+
+const SCRIPT_OPTIONS: {
```
danixeee: here you also repeat some things, maybe it can be combined as well

plainly-plugin/src/ui/components/parametrization/utils.ts
danixeee, line 1: in this file also, a lot of those maps `[scriptType]: something`, maybe we can align this somehow

plainly-plugin/src/ui/components/parametrization/LayerManagementScriptDialog.tsx
vladavoX: to utils.ts, and will rename it to something like `DYNAMIC_SCRIPT_...` so there is no confusion, this is for that only

plainly-plugin/src/ui/components/parametrization/ParametrizedLayers.tsx
vladavoX: yes, that is the reason. We can remove the `type as unknown as string` by making the thing above `Partial<Record<string, string>>`

and this thing `raw.charAt(0).toUpperCase() + raw.slice(1).toLowerCase().replace(/_/g, ' ')` is just a safe guard in case of a non existing script above, so we make the text look good

plainly-plugin/src/node/request.ts
```diff
@@ -61,11 +61,19 @@ async function get<T>(
 async function post<T>(
   path: string,
   apiKey: string,
-  body: string,
+  body: unknown,
```
danixeee: use `body: T`

danixeee (changes requested):
Make these consistent, add header

Image

Image

Filter still have bugs:

https://github.com/user-attachments/assets/4debc56a-7ed3-443e-bdc9-7e237798e466

plainly-plugin/src/node/request.ts
vladavoX: not `T`, a new one `B`, like in `put`, `T` is for response.

danixeee (approved):

PR #147, opened 2026-03-31
Layout improvements

vladavoX:
## Summary

There are two issues with #137 . 
1. First is that Sidebar is now getting small, because new link is named `Parametrization`, and we had fixed width of `36. This is solved by getting real time data of sidebar width.
2. Second is that some z values are wrong, this was seen in new `sticky` element in a Parametrized layers list, it is visible while dialog is visible also. So I've adjusted some `z`s in this order:
    - max z -> tooltip and reload button
    - then notification overlay
    - then dialogs and sidebar
    - then pin + backdrop blurs
    - then these stickies and menus / dropdowns we have
3. Tooltip refactor -> use BaseUI tooltip, shadcn guide for manual implementation

## Type of change

- [x] Bug fix
- [ ] New feature
- [x] Refactor / internal improvement
- [ ] Dependency update
- [ ] Docs / tooling

## Related issue

#137 

## Testing

- [ ] Existing tests pass (`yarn test`)

**OS tested on:**

## Checklist

### DIFF
--- a/plainly-plugin/package.json
+++ b/plainly-plugin/package.json
@@ -13,6 +13,7 @@
     "test:ci:coverage": "jest --ci --coverage"
   },
   "dependencies": {
+    "@base-ui/react": "^1.3.0",
     "@dnd-kit/core": "^6.3.1",
     "@dnd-kit/sortable": "^8.0.0",
     "@headlessui/react": "^2.2.0",
@@ -21,6 +22,7 @@
     "archiver": "^5.3.2",
     "axios": "^1.13.5",
     "classnames": "^2.5.1",
+    "clsx": "^2.1.1",
     "date-fns": "^4.1.0",
     "form-data": "^4.0.4",
     "lodash-es": "^4.17.23",
@@ -28,7 +30,8 @@
     "react": "^18.3.1",
     "react-dom": "^18.3.1",
     "react-hooks-global-state": "^2.1.0",
-    "semver": "^7.7.4"
+    "semver": "^7.7.4",
+    "tailwind-merge": "^3.5.0"
   },
   "devDependencies": {
     "@babel/core": "^7.28.6",
--- a/plainly-plugin/src/ui/App.tsx
+++ b/plainly-plugin/src/ui/App.tsx
@@ -1,6 +1,12 @@
 import { isDev, pluginBundleVersion } from '@src/env';
 import { useMemo } from 'react';
-import { Banner, Button, ExternalLink, Sidebar } from './components';
+import {
+  Banner,
+  Button,
+  ExternalLink,
+  Sidebar,
+  TooltipProvider,
+} from './components';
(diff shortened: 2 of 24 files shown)


vladavoX (commented):
## Code Review

### Overview
Fixes two layout regressions from #137: sidebar width being too narrow for the new "Parametrization" nav link, and incorrect z-index stacking causing sticky elements to bleed through dialogs.

---

### What's Good

**ResizeObserver for dynamic sidebar width** — The CSS custom property approach (`--sidebar-width`) is clean and elegant. Setting it on `document.documentElement` makes it globally available, and replacing the hardcoded `xs:w-36` with `xs:w-max` + observer correctly handles variable-length link labels.

**z-index hierarchy** — The new layering is clearly reasoned and consistently applied across all affected files:
- z-50: tooltip + reload button
- z-40: notifications, dropdowns
- z-30: sidebar + dialog content
- z-20: dialog backdrops, pin/API key overlays

All dialog components (`ConfirmationDialog`, `ScriptDialogShell`, `ScriptsDialog`) were updated consistently. No missed `ml-36` hardcodes remain.

---

### Suggestions

**Minor: prefer `entry.contentRect.width` over `el.offsetWidth`**

Reading `offsetWidth` inside a ResizeObserver callback triggers an extra layout reflow. Since the observer already provides the new dimensions in the `entries` argument, you can avoid this:

```ts
const observer = new ResizeObserver((entries) => {
  const width = entries[0]?.contentRect.width;
  if (width != null) {
    document.documentElement.style.setProperty('--sidebar-width', `${width}px`);
  }
});
```

This is a minor point since the sidebar rarely resizes, but it's the idiomatic usage.

**Minor: initial CSS variable value**

On first mount, `--sidebar-width` is unset until the ResizeObserver fires (after the first paint). In practice this is imperceptible, but if you want belt-and-suspenders you could set it once after `ref.current` is confirmed:

```ts
useEffect(() => {
  const el = ref.current;
  if (!el) return;
  // Set immediately so the first render is correct
  document.documentElement.style.setProperty('--sidebar-width', `${el.offsetWidth}px`);
  const observer = new ResizeObserver((entries) => {
    const width = entries[0]?.contentRect.width;
    if (width != null) document.documentElement.style.setProperty('--sidebar-width', `${width}px`);
  });
  observer.observe(el);
  return () => observer.disconnect();
}, []);
```

**Worth noting: `w-max` is unbounded**

The sidebar now grows to fit its content. That's correct for the current use case, but if a nav link label ever becomes very long (or a label is dynamic), the sidebar could get unexpectedly wide. Not a blocker — just worth keeping in mind if labels are ever data-driven.

---

### Verdict

The approach is sound and the changes are consistently applied. The two minor suggestions above are optional improvements, not blockers. 

🟢 Approved with optional nits.

danixeee (approved):

PR #150, opened 2026-04-10
relates to #4647: enable parameter edit

vladavoX:
## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Dependency update
- [ ] Docs / tooling

## Testing

**OS tested on:**

## Checklist

### DIFF
--- a/plainly-plugin/src/ui/components/common/Checkbox.tsx
+++ b/plainly-plugin/src/ui/components/common/Checkbox.tsx
@@ -6,6 +6,7 @@ export function Checkbox({
   label,
   description,
   defaultChecked = true,
+  checked,
   onChange,
   disabled = false,
 }: {
@@ -14,9 +15,12 @@ export function Checkbox({
   label: string;
   description?: string;
   defaultChecked?: boolean;
+  checked?: boolean;
   onChange: React.Dispatch<React.SetStateAction<boolean>>;
   disabled?: boolean;
 }) {
+  const checkProps = checked === undefined ? { defaultChecked } : { checked };
+
   return (
     <div
       className={classNames(
@@ -27,7 +31,7 @@ export function Checkbox({
       <div className="flex h-6 shrink-0 items-center">
         <div className="group grid size-4 grid-cols-1">
           <input
-            defaultChecked={defaultChecked}
+            {...checkProps}
             id={id}
             name={name}
             type="checkbox"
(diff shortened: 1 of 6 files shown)


vladavoX (commented):
## Review

### Overview

Implements parameter editing for parametrized layers. Key additions:
- **`ParameterDialog`** — new modal for editing static/dynamic parameter settings (value, parameter name with regex validation, default value, mandatory flag)
- **`SortableLayerItem`** — layer rows extracted into a sortable component, enabling layer drag-and-drop reordering
- **`Checkbox`** — upgraded to support controlled (`checked`) alongside uncontrolled (`defaultChecked`)
- **`ensureHashPrefix` / `stripHashPrefix`** — symmetric util helpers for `#`-prefixed values
- **`LayerBase.label`** — new required field on the type

---

### Issues

**Bug — Static save retains stale dynamic fields** (`ParameterDialog.tsx:158-170`)

When saving in Static mode, the spread carries over `mandatory` and `defaultValue` from the previous Dynamic configuration:
```ts
: {
    ...currentParametrization,   // ← keeps mandatory, defaultValue from dynamic mode
    expression: false,
    value: staticValue,
  };
```
Any downstream code reading `mandatory` or `defaultValue` without first checking `expression === true` will see stale data. These should be explicitly cleared in the static branch.

---

**Edit button not respecting `disabled` prop** (`ParametrizedLayers.tsx:566-572`)

The Edit button has no `disabled` attribute, so users can open the parameter dialog even when the parent's `disabled` prop is true. The drag handle correctly checks `isFilterActive` — the edit button should too.

---

**`PARAM_NAME_REGEX` defined locally instead of in `utils.ts`** (`ParameterDialog.tsx:112`)

`SCRIPT_PARAMETER_NAME_REGEX` lives in `utils.ts`. `PARAM_NAME_REGEX` should also live there so both are in one place and accessible to any future validation logic.

---

**`defaultV` is an opaque variable name** (`ParameterDialog.tsx:139`)

`defaultV` is unclear at a glance. Rename to `paramDefaultValue` or similar — `defaultValue` is already destructured above, so a distinct but readable name avoids both shadowing and the abbreviation.

---

### Code Quality

- **`checkProps` spread in Checkbox** — clean solution to the uncontrolled/controlled switch without duplicating the `<input>`.
- **`key` trick on `ParameterDialog`** (`ParametrizedLayers.tsx:865`) — correct pattern to reset dialog state on open/close; intentional and clear.
- **`handleLayerDragEnd` clears selection** — reasonable UX; drag reorders so prior selection is ambiguous.
- **`isFilterActive` computation** correctly gates drag when `disabled`, filter, or type query is active.
- **State initialization in `ParameterDialog`** — relies on `key` prop to reset when reopened for a different layer; valid approach.

PR #155, opened 2026-04-22
relates to #4647: premade script Shift and Crop

vladavoX:
## Summary

1. ✅ Handle adding a premade script Shift and Crop to type of render comp that has:
    - Intro comp
    - Scene comps -> each comp contains a video
    - Outro
2. ✅ Handle adding a premade script Shift and Crop to type of render comp that has:
    - Intro comp
    - Single video comp -> that has all the videos
    - Outro comp
3. ✅ Handle adding a premade script Shift and Crop to type of render comp that has:
    - Intro comp
    - Videos directly in render comp (no comp for videos)
    - Outro comp
4. ✅ overlap
5. ✅ single script add
    
Project used for testing [ShiftInCrop Combo.zip](https://github.com/user-attachments/files/26972772/ShiftInCrop.Combo.zip)

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Dependency update
- [ ] Docs / tooling

## Testing

**OS tested on:**

## Checklist


vladavoX:
## Overview

Adds a "Shift and Crop" premade script action that chains `SHIFT_IN` scripts across selected scene layers and applies a `CROP` to the last scene, with special handling for three comp layouts (per-scene comps, single-video comp, direct-video render comp). Adds three new ExtendScript utilities (`getSelectedLayers`, `getAllVideoLayersInComp`, `getAllAudioLayersInComp`), a `ChoiceDialog`, a `PremadeScriptsDialog`, and a `PremadeScriptRegistry`.

## Issues

### Correctness / Type Safety

- **`label: string` is added as required on `LayerBase`** (`types/template.ts:12`) but is only populated in the new shift/crop code. Layers created elsewhere (e.g. returned from the backend / mapped from `ProjectData` in Plainly API responses) and existing persisted templates won't include it. Nothing in the codebase actually *reads* `layer.label`. Either (a) make it optional (`label?: string`) and remove it from the synthesized objects, or (b) populate it everywhere layers originate. As written, it is dead weight plus a type hazard.

> Fixed

- ~**Case 3 mismatch with PR description.** The PR summary says case 3 (videos directly in render comp) is not handled, but the code *does* have an `isDirectVideo`/`isDirectAudio` path. Either update the PR description or remove the dead branch. If it's meant to work, the case should be tested and checked off.~

> Ignore

- **`Number(sceneLayer.internalId)` silently becomes `NaN`** if `internalId` isn't numeric. `itemByID(NaN)` will return `null` and we'll just return an empty layer list — fine, but worth a tiny guard or an early notify.

> Fixed

- **Audio/video overlap list.** `VIDEO_FILE_EXTENSIONS` and `AUDIO_FILE_EXTENSIONS` both contain `mpg`, `mpa`, `mpe`, `webm`. Because `hasAudioExtension` short-circuits on `hasVideoExtension`, those are always classified as video. That's consistent with the comment, but audio-only `.webm` exists in the wild and will be misclassified. At minimum, drop the duplicates from the audio list so readers don't assume they'd ever match.

> Fixed

### UX / Behavior

- **Single prompt for all ambiguous scenes.** If multiple comps have both audio + video inner layers, the user gets one prompt and the choice applies to all of them. The description lists the affected comps so this is discoverable, but it prevents mixed video/audio chains. Worth confirming this matches product intent (or call it out in the PR).

> Agreed to have a single prompt

- **`ChoiceDialog` and `PremadeScriptsDialog` duplicate ~50 lines of styling** (`DialogBackdrop` + `DialogPanel` + sidebar-aware `ml-*` classes). An existing `ConfirmationDialog` probably has the same shell. Consider extracting a `DialogShell` so the three share a backdrop/panel. Not a blocker.

> Different PR, less changes now

- ~**Stale-closure risk on async await.** `shiftAndCropHandler` reads `editableLayers` before `await promptChoice(...)`, then only uses it to look up `existing` entries. The final `setEditableLayers` uses the functional form (good). In practice state merge works out, but if the user edits layers during the prompt, the `existing ?? sceneLayer` lookup uses stale data. Low impact; worth a comment acknowledging it.~

> Ignore

### ExtendScript

- `getSelectedLayers` / `getAllVideoLayersInComp` / `getAllAudioLayersInComp` all return `JSON.stringify([])` on failure, which the TS side parses to `[]`. That's the right contract, but there's no signal distinguishing "no active comp" from "comp has no videos". Callers can't tell the difference; consider whether that matters for future error messages.

> Fixed

- ~In `AeScriptsApi.ts`, wrapping a `number` compId with `JSON.stringify(String(compId))` before embedding in evalScript is belt-and-suspenders but harmless. Leave it.~

> Ignore

### Process / PR Hygiene

- ~All checklist boxes are unchecked: no confirmation of manual AE testing, `yarn biome`, `yarn build`, or TS cleanness. Given this PR touches both `plainly-types` (public types) and the plugin UI/bridge, those checks are load-bearing before merge.~

> Done

- ~PR is still Draft — looks ready apart from the `label` type issue and the case-3 documentation mismatch.~

> Not in Draft anymore

## Summary

The core design is clean: the registry pattern is nice and extensible, `upsertScript` is a good utility, and the scene/inner-media separation is readable. The main blocker is the non-optional `label` field on `LayerBase`, which should either be removed/optionalized or populated at every layer origin. Secondary: reconcile case 3 between code and PR description, dedupe audio/video extension lists, and run the checklist commands.

plainly-aescripts/src/utils.ts
```diff
@@ -241,6 +377,136 @@ function selectFile(fileId: string): void {
...
+          info.isSolid = true;
+        } else if (src.file != null) {
+          const fsName = src.file.fsName;
+          if (hasVideoExtension(fsName)) {
+            info.isVideo = true;
+          } else if (hasAudioExtension(fsName)) {
+            info.isAudio = true;
+          } else if (hasImageExtension(fsName)) {
+            info.isImage = true;
+          }
```
danixeee: i think you do not need these functions here with file extensions, you can check it like this:

```ts
const isImage = (src: FootageItem): boolean => {
    return src.mainSource.isStill;
}
const isVideo = (src: FootageItem): boolean => {
    return src.hasVideo;
}
const isAudio = (src: FootageItem): boolean => {
    return src.hasAudio && !src.hasVideo;
}
```

plainly-plugin/src/ui/components/parametrization/premadeScripts/shiftAndCrop.ts
```diff
@@ -0,0 +1,385 @@
...
+
+export const shiftAndCropHandler: PremadeScriptHandler = async ({
+  editableLayers,
+  setEditableLayers,
+  notifyError,
+  notifyInfo,
+  notifySuccess,
+  promptChoice,
+}) => {
+  let selected: Awaited<ReturnType<typeof AeScriptsApi.getSelectedLayers>>;
```
danixeee: what is this?

plainly-plugin/src/ui/components/parametrization/premadeScripts/shiftAndCrop.ts
```diff
@@ -0,0 +1,385 @@
...
+ * Returns 0 when the layers are gapped or perfectly butted. Both layers must
+ * live in the same parent comp (so the same frame rate applies).
+ */
+const overlapFrames = (
+  prev: { outPoint: number },
+  curr: { inPoint: number; compFrameRate: number },
+): number => {
+  const seconds = prev.outPoint - curr.inPoint;
+  if (seconds <= 0) return 0;
+  return Math.max(0, Math.round(seconds * curr.compFrameRate));
```
danixeee: you have `if (seconds <= 0) return 0;` so you don't need `Math.max`

plainly-plugin/src/ui/components/parametrization/premadeScripts/shiftAndCrop.ts
```diff
@@ -0,0 +1,385 @@
...
+        l.layerName === sel.name,
+    );
+    const sceneLayer: Layer = existing ?? {
+      internalId: String(sel.sourceCompId),
+      layerName: sel.name,
+      label: sel.name,
+      compositions: [{ id: sel.compId, name: sel.compName }],
+      layerType: 'COMPOSITION',
+    };
+    return { sel, sceneLayer, directMediaType: null as MediaType | null };
```
danixeee: this is weird here `directMediaType: null as MediaType | null`

danixeee (commented):
Image

why we do not add these scripts to parameters that already exist in this list? Also, it would be cool if it's possible to mark new scripts other colour, until we save changes, so we can see what is added, etc.. Probably not in this PR, just UX feedback

Also, let claude review this PR one more time

vladavoX:
@danixeee I've adressed your comments, and ran multiple agent review with `xhgih` effort, found some issues, and fixed them.

We can make a call for the creation of duplicate parameter issue you had, couldn't replicate

PR #158, opened 2026-04-24
fix: clear stale PIN on api key save and invalid decode

vladavoX:
## Summary
- Clear the sessionStorage PIN after a successful `setSettingsApiKey` so changing the PIN re-prompts the overlay on the next protected route (previously the stale PIN was reused, causing an "Invalid PIN entered" crash).
- In `AuthProvider`, decode the API key inside a try/catch before rendering and clear the PIN via `useEffect` when decode fails, so any stale/corrupted PIN falls back to the overlay instead of throwing during render.

## Repro (before fix)
1. Save API key with PIN `1337`.
2. Visit a protected route, unlock with `1337`.
3. Go back to Settings, change PIN to `1111`, save.
4. Visit a protected route → app crashes with "Invalid PIN entered".

After fix: step 4 shows the PIN overlay asking for the new PIN.

## Test plan
- [x] Save API key with PIN, unlock overlay, change PIN in settings, navigate to a protected route → overlay re-prompts with new PIN.
- [x] Save API key without PIN → no overlay, direct access.
- [x] Save API key with PIN, then re-save API key without PIN → no overlay, direct access (stale PIN cleared).
- [x] Manually corrupt `sessionStorage.pin` to an invalid value and reload a protected route → overlay appears instead of crashing.
- [x] Normal unlock flow still works.

### DIFF
--- a/plainly-plugin/src/ui/components/settings/AuthProvider.tsx
+++ b/plainly-plugin/src/ui/components/settings/AuthProvider.tsx
@@ -3,8 +3,8 @@ import {
   useSessionStorage,
   useSettings,
 } from '@src/ui/hooks';
-import { LoaderCircleIcon } from 'lucide-react';
 import { createContext, useCallback, useEffect } from 'react';
+import { Loading } from '../common';
 import { MissingApiKey, PinOverlay } from '.';
 
 interface AuthContextProps {
@@ -38,11 +38,11 @@ export const AuthProvider = ({ children }: { children: React.ReactNode }) => {
   );
 
   // Try to decode BEFORE returning JSX. Don't throw, just track failure.
-  let decryptedApiKey: string | undefined;
+  let settingsApiKey: string | undefined;
   let pinInvalid = false;
   if (apiKeySet && (!apiKeyEncrypted || pin)) {
     try {
-      decryptedApiKey = getSettingsApiKey(pin);
+      settingsApiKey = getSettingsApiKey(pin);
     } catch {
       pinInvalid = true;
     }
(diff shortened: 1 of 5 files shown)


plainly-plugin/src/ui/components/settings/AuthProvider.tsx
```diff
@@ -38,7 +37,24 @@ export const AuthProvider = ({ children }: { children: React.ReactNode }) => {
     [getSettingsApiKey, notifyError, setPinStorage],
   );
 
-  const showOverlay = apiKeyEncrypted && !pin;
+  // Try to decode BEFORE returning JSX. Don't throw, just track failure.
+  let decryptedApiKey: string | undefined;
+  let pinInvalid = false;
+  if (apiKeySet && (!apiKeyEncrypted || pin)) {
```
danixeee: can you just early return if apiKeySet is false? to make the logic more clear.

`if (apiKeySet) { return; }` ?

plainly-plugin/src/ui/components/settings/AuthProvider.tsx
```diff
@@ -38,7 +37,24 @@ export const AuthProvider = ({ children }: { children: React.ReactNode }) => {
...
+    }
+  }
+
+  // If decode failed, clear the PIN on the next tick.
+  // This triggers a re-render where `pin` is now `undefined`, showing the overlay again.
+  useEffect(() => {
+    if (pinInvalid) clearPinStorage();
+  }, [clearPinStorage, pinInvalid]);
+
+  const showOverlay = apiKeyEncrypted && (!pin || pinInvalid);
```
danixeee: this is also a bit confusing

danixeee (commented):
Very confusing IMO. In order to change API key, you need to know the old one right? I would say when someone change it, just override session storage (do not clear it). This should remove many changes if I understand how this works.. Wdyt?

plainly-plugin/src/ui/components/settings/AuthProvider.tsx
vladavoX: i can't return, I need to render something (MIssingApiKey component)

plainly-plugin/src/ui/components/settings/AuthProvider.tsx
vladavoX: why? We show overlay if `apiKeyEncrypted` and there is no `pin` or `pinInvalid`?

plainly-plugin/src/ui/components/settings/AuthProvider.tsx
```diff
@@ -16,51 +16,62 @@ export const AuthContext = createContext<AuthContextProps>(
...
+      }
+    },
+    [getSettingsApiKey],
+  );
+
+  const onPinSubmitted = useCallback(
+    (submittedPin: string | undefined) => {
+      const { error } = tryUnlock(submittedPin);
+      if (error) {
         notifyError('There was an issue with setting the PIN.', error);
```
danixeee: setting the PIN, is it the right message here?

danixeee (approved):

PR #161, opened 2026-04-27
feat: add folder field to upload form

### LINKED ISSUE #114: Include folder field in upload project
Missing field `folder` from project upload

Image

vladavoX:
## Summary

Adds folder input same as on Plainly platform, to add or choose existing folder

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Dependency update
- [ ] Docs / tooling

## Related issue

closes #114 

## Testing

**After Effects version tested:** 26
**OS tested on:** MacOS

## Screenshots

image
image

## Checklist

### DIFF
--- a/plainly-plugin/src/ui/components/form/FolderPicker.tsx
+++ b/plainly-plugin/src/ui/components/form/FolderPicker.tsx
@@ -1,0 +1,209 @@
+import {
+  Dialog,
+  DialogBackdrop,
+  DialogPanel,
+  DialogTitle,
+} from '@headlessui/react';
+import { useGetProjects } from '@src/ui/hooks/api';
+import { State, useGlobalState } from '@src/ui/state/store';
+import classNames from 'classnames';
+import { FolderIcon, LoaderCircleIcon, XIcon } from 'lucide-react';
+import { useMemo, useState } from 'react';
+import { Button } from '../common';
+import { Description, Label } from '../typography';
+
+export function FolderPicker({
+  folder,
+  onChange,
+}: {
+  folder?: string;
+  onChange: (e: React.ChangeEvent<HTMLInputElement>) => void;
+}) {
+  const [dialogOpen, setDialogOpen] = useState(false);
+
+  const dispatch = (value: string) => {
+    onChange({
+      target: { name: 'folder', value },
+    } as React.ChangeEvent<HTMLInputElement>);
+  };
+
+  return (
+    <div className="col-span-full">
+      <Label label="Folder" htmlFor="folder" />
+      <Description>
+        Folders help you organize your projects on Plainly Videos.
+      </Description>
+      <div className="flex mt-1 col-start-1 row-start-1 w-full">
+        <div className="relative w-full">
+          <input
+            id="folder"
+            name="folder"
+            type="text"
+            className="w-full rounded-l-md bg-white/5 px-3 py-1 text-xs text-white outline outline-1 -outline-offset-1 outline-white/10 focus-within:outline focus-within:outline-2 focus-within:-outline-offset-2 focus-within:outline-indigo-500 disabled:opacity-50 disabled:cursor-not-allowed"
+            value={folder === '/' ? 'Root' : folder || 'No folder'}
+            disabled
+          />
+          {folder && folder !== '/' && (
+            <XIcon
+              className="size-4 absolute right-2 top-1/2 -translate-y-1/2 text-gray-400 cursor-pointer hover:text-gray-200"
+              onClick={() => dispatch('/')}
+            />
+          )}
+        </div>
+        <button
+          type="button"
+          className="-ml-px rounded-r-md bg-white/10 hover:bg-white/20 px-3 py-1 text-xs text-white outline outline-1 -outline-offset-1 outline-white/10 focus-within:outline focus-within:outline-2 focus-within:-outline-offset-2 focus-within:outline-indigo-500 whitespace-nowrap"
+          onClick={() => setDialogOpen(true)}
+        >
+          Choose folder
+        </button>
+      </div>
+      <FolderPickerDialog
+        open={dialogOpen}
+        onClose={() => setDialogOpen(false)}
+        onSelect={(value) => {
+          dispatch(value);
+          setDialogOpen(false);
+        }}
+      />
+    </div>
+  );
+}
+
+function FolderPickerDialog({
+  open,
+  onClose,
+  onSelect,
+}: {
+  open: boolean;
+  onClose: () => void;
+  onSelect: (value: string) => void;
+}) {
+  const { data, isLoading } = useGetProjects();
+  const [settings] = useGlobalState(State.SETTINGS);
+  const sidebarOpen = settings.sidebarOpen;
+
+  const [folderName, setFolderName] = useState('');
+
+  const allFolders = useMemo(() => {
+    const folders = new Set<string>();
+
+    data?.forEach((p) => {
+      const folderPath = p.attributes?.folder;
+      if (typeof folderPath === 'string' && folderPath.trim() !== '') {
+        const parts = folderPath.split('/');
+        for (let i = 1; i <= parts.length; i++) {
+          const partialPath = parts.slice(0, i).join('/');
+          if (partialPath) folders.add(partialPath);
+        }
+      }
+    });
+
+    return Array.from(folders).sort();
+  }, [data]);
+
+  return (
+    <Dialog open={open} onClose={onClose} className="relative">
+      <DialogBackdrop
+        transition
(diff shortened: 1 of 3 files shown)


plainly-plugin/src/ui/components/form/UploadForm.tsx
```diff
@@ -113,6 +122,11 @@ export function UploadForm() {
         }
       }
 
+      const folder = inputs.folder?.trim();
+      if (folder !== undefined && folder !== null) {
```
vladavoX: `inputs.folder?.trim()` returns `string | undefined` — it can never be `null`, so the `!== null` check is dead.

More importantly: an empty string still satisfies `folder !== undefined`, so if the user types into the folder field and then deletes it, you end up appending an empty `folder` value to the FormData. Did you mean something like:

```ts
if (folder) {
  formData.append('folder', folder);
}
```

plainly-plugin/src/ui/components/form/UploadForm.tsx
```diff
@@ -181,6 +195,10 @@ export function UploadForm() {
       if (name === 'projectName' || name === 'description') {
         setInputs((prev) => ({ ...prev, [name]: value }));
       }
+
+      if (name === 'folder') {
+        setInputs((prev) => ({ ...prev, folder: value }));
```
vladavoX: This branch is structurally identical to the `projectName | description` one above — could be folded in:

```ts
if (name === 'projectName' || name === 'description' || name === 'folder') {
  setInputs((prev) => ({ ...prev, [name]: value }));
}
```

Minor, but avoids the duplicate `setInputs` call.

plainly-plugin/src/ui/components/form/UploadForm.tsx
```diff
@@ -337,15 +357,49 @@ function Inputs({
...
 }: {
   projectName?: string;
   description?: string;
   tags?: string[];
+  folder?: string;
   onChange: (
     e: React.ChangeEvent<HTMLInputElement | HTMLTextAreaElement>,
   ) => void;
 }) {
+  const { data, isLoading } = useGetProjects();
```
vladavoX: `useGetProjects` is called unconditionally inside `Inputs`, so the full project list is fetched (and polled, since it has a `refetchInterval`) every time the upload form renders — even when the user never opens the folder dialog.

Consider moving this hook into a child component that only mounts when `folderDialogOpen` is true, so the request only fires when the picker is actually used.

plainly-plugin/src/ui/components/form/UploadForm.tsx
```diff
@@ -382,12 +436,155 @@ function Inputs({
...
+        </Description>
+        <div className="flex mt-1 col-start-1 row-start-1 w-full">
+          <div className="relative w-full">
+            <input
+              id="folder"
+              name="folder"
+              type="text"
+              className="w-full rounded-l-md bg-white/5 px-3 py-1 text-xs text-white outline outline-1 -outline-offset-1 outline-white/10 focus-within:outline focus-within:outline-2 focus-within:-outline-offset-2 focus-within:outline-indigo-500 disabled:opacity-50 disabled:cursor-not-allowed"
+              value={folder === '/' ? 'Root' : folder || 'No folder'}
+              onChange={onChange}
```
vladavoX: Disabled inputs don't fire change events, so this `onChange={onChange}` is dead. The displayed value is also derived from `folder` and pretty-printed (`'Root'` / `'No folder'`), so there's nothing meaningful for `onChange` to forward even if it did fire — safe to drop.

plainly-plugin/src/ui/components/form/UploadForm.tsx
```diff
@@ -382,12 +436,155 @@ function Inputs({
...
+              disabled
+            />
+            {folder && folder !== '/' && (
+              <XIcon
+                className="size-4 absolute right-2 top-1/2 -translate-y-1/2 text-gray-400 cursor-pointer hover:text-gray-200"
+                onClick={() => {
+                  onChange({
+                    target: {
+                      name: 'folder',
+                      value: '/',
```
vladavoX: The clear (X) button sets `folder` to `'/'`, which then gets sent as `folder=/` on submit. If the goal is "unset the folder", clearing back to `undefined` (so the field is omitted from the FormData) might match the API better. Worth confirming against the platform's semantics — `'/'` may or may not mean "root" vs. "no folder" on the server side.

plainly-plugin/src/ui/components/form/UploadForm.tsx
```diff
@@ -382,12 +436,155 @@ function Inputs({
...
+                  secondary
+                  className="inline-flex mt-2 sm:mt-0 w-full sm:w-auto justify-center"
+                >
+                  Cancel
+                </Button>
+              </div>
+            </DialogPanel>
+          </div>
+        </div>
+      </Dialog>
```
vladavoX: Optional: the dialog (its state, the projects fetch, the folder aggregation, and the JSX) is ~150 lines and quite self-contained. Extracting it into a `<FolderPicker value={folder} onSelect={...} />` would keep `Inputs` focused on form fields and make the picker reusable if you ever need it on the project list / detail page too.

vladavoX (commented):
## Summary

Adds a folder picker to the upload form: a read-only display field with a clear (X) button, plus a dialog that lets the user type a path or pick from existing folders aggregated from `useGetProjects()`. The `Project` type gains an optional `folder` attribute, which is forwarded into the upload `FormData`.

Functionality matches the screenshots and looks right end-to-end. A few correctness/clarity nits inline. Two more substantive points worth a look:

- `useGetProjects()` is called (and polled, via its `refetchInterval`) eagerly inside `Inputs`, even when the dialog is never opened — easy to gate behind dialog open.
- The clear (X) button resets to `'/'` rather than `undefined` — double-check that matches the platform's idea of "no folder".

_Side note: the `mt-2` → `mt-1` change on the Tags input is unrelated to this PR — fine to keep, just flagging._

Nothing blocking from my side.

plainly-plugin/src/ui/components/form/UploadForm.tsx
vladavoX: no, it is good, `X` removes it, and sends "/", same what we have on our platform

plainly-plugin/src/ui/components/form/FolderPicker.tsx
```diff
@@ -0,0 +1,209 @@
...
+function FolderPickerDialog({
+  open,
+  onClose,
+  onSelect,
+}: {
+  open: boolean;
+  onClose: () => void;
+  onSelect: (value: string) => void;
+}) {
+  const { data, isLoading } = useGetProjects();
```
danixeee, line 82: is this same on UI, calling get projects to find folders?

plainly-plugin/src/ui/components/form/FolderPicker.tsx
```diff
@@ -0,0 +1,209 @@
...
+  const [settings] = useGlobalState(State.SETTINGS);
+  const sidebarOpen = settings.sidebarOpen;
+
+  const [folderName, setFolderName] = useState('');
+
+  const allFolders = useMemo(() => {
```
danixeee, line 88: maybe you can create `useGetProjectsFolders` and move this logic to a hook?

plainly-plugin/src/ui/components/form/FolderPicker.tsx
```diff
@@ -0,0 +1,209 @@
...
+                  <input
+                    id="folderName"
+                    name="folderName"
+                    type="text"
+                    className="col-start-1 row-start-1 block w-full rounded-md bg-white/5 px-3 py-1 text-xs text-white outline outline-1 -outline-offset-1 outline-white/10 placeholder:text-gray-500 focus:outline focus:outline-2 focus:-outline-offset-2 focus:outline-indigo-500"
+                    placeholder="/folder/sub-folder"
+                    value={folderName}
+                    onChange={(e) => {
+                      const value = e.target.value.trim();
+                      if (value && !value.startsWith('/'))
```
danixeee, line 151: I thought biome is configured to force { } in these cases? I think it's better to use them everywhere

danixeee (approved):

plainly-plugin/src/ui/components/form/FolderPicker.tsx
vladavoX: imo only if we are going to use it in projects list, if not, I would leave it as it is

plainly-plugin/src/ui/components/form/FolderPicker.tsx
vladavoX: it is not, will do quick PR for that

plainly-plugin/src/ui/components/form/FolderPicker.tsx
vladavoX: but then, I would add it then, no need to do anything now

PR #162, opened 2026-04-28
fix: remove dead code

vladavoX:
## Summary

Removes dead code across the plugin

## Type of change

- [ ] Bug fix
- [ ] New feature
- [x] Refactor / internal improvement
- [ ] Dependency update
- [ ] Docs / tooling

## Testing

- [ ] Tested manually in After Effects
- [ ] Existing tests pass (`yarn test`)

**OS tested on:**

## Checklist

### DIFF
--- a/plainly-aescripts/src/utils.ts
+++ b/plainly-aescripts/src/utils.ts
@@ -247,7 +247,6 @@ export {
   getInstalledFontsByPostScriptName,
   getTextLayersByComp,
   isWin,
-  pathJoin,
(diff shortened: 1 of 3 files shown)


danixeee (approved):
What happened with these deps?

vladavoX:
> What happened with these deps?

- heroicons, we use lucide
- globals, that was something for jest, but it works without it
- typography, pulled same as frontend, but we don't use that

PR #171, opened 2026-09-14
fix: preserve image sequences on export

vladavoX:
## Problem

Image sequences were collapsed into a single frame whenever a project was exported or uploaded, instead of staying as authored in the project.

## Root cause

Three points in the collect → copy → relink pipeline only ever handled the first frame:

- `collectFootage` (`plainly-aescripts/src/collect.ts`) recorded `item.file`, which for a sequence is only the first frame
- `copyFootage` (`plainly-plugin/src/node/collectProject/copyFootage.ts`) copied that single file into `(Footage)`
- `relinkFootage` (`plainly-aescripts/src/relink.ts`) called `FootageItem.replace()`, which always imports a still — this alone collapses the sequence even with the other two fixed

## Changes

- `Footage` gains `isSequence`, detected from a non still `FileSource` pointing at a still image extension. After Effects exposes no direct flag, and video containers are also non still, so the extension check does the discrimination. PSD and AI are left out since relinking already skips them.
- `RelinkData` values become `{ path, isSequence }` instead of a bare path string.
- `copyFootage` copies every frame of a sequence, matched from the trailing frame number in the first frame's name (`shot_0001.png` → `^shot_\d+\.png$`). A non numbered name falls back to a single file copy.
- `relinkFootage` uses `replaceWithSequence(file, false)` for sequences — numbered order, not alphabetical. This covers the undo path too, so restoring the original links keeps the sequence intact.

## Known limitation

All frames matching the sequence naming in the source folder are copied, not only the range the project actually uses. That can make the zip larger than strictly needed; the render is unaffected. Narrowing it needs `item.duration / item.frameDuration` math and was left out for now.

## Testing

Not yet verified in After Effects — needs a manual run: export a project containing an image sequence, unzip, and confirm the frame count and that the relinked footage is still a sequence.


plainly-aescripts/src/collect.ts
```diff
@@ -61,6 +61,82 @@ function collectFonts(): Font[] {
...
+  'nef',
+  'arw',
+  'raf',
+  'orf',
+];
+
+/**
+ * Determines whether a footage item is an image sequence rather than a single file.
+ *
+ * After Effects exposes no direct flag for this, so it is derived from the source
```
danixeee, line 108: @vladavoX can you please double check if this is true

see here: https://ae-scripting.docsforadobe.dev/

plainly-aescripts/src/collect.ts
```diff
@@ -61,6 +61,82 @@ function collectFonts(): Font[] {
...
+    return false;
+  }
+
+  const name = item.file.name.toLowerCase();
+  const dotIndex = name.lastIndexOf('.');
+  if (dotIndex === -1) {
+    return false;
+  }
+
+  const extension = name.substring(dotIndex + 1);
```
danixeee, line 130: what this does? only checks file extension? are we sure this is correct

plainly-plugin/src/node/collectProject/copyFootage.ts
```diff
@@ -8,6 +8,72 @@ import {
   runInParallelReturnRejected,
 } from '../utils';
 
+function escapeRegExp(value: string): string {
```
danixeee, line 11: where is this used and why we need it?

plainly-aescripts/src/collect.ts
vladavoX: there is no way to check if it is image sequence directly

plainly-aescripts/src/collect.ts
vladavoX: yes, because `isStill` returns videos too, and we need the extension to then check if it is in image format

plainly-plugin/src/node/collectProject/copyFootage.ts
vladavoX: so we can get files like `image.v1_0001.png` for example, and similar. We can do `startsWith/slice`, but imo, it doesn't matter?

danixeee (approved):

Issue #173, opened 2026-09-18
Codebase audit: security and correctness findings


vladavoX:
Static audit of `master` (34dd862). Findings grouped by severity; each item cites the line it anchors to. Nothing has been changed yet.

## Security

1. **PIN protection is effectively cosmetic.** `plainly-plugin/src/node/encoding.ts:4-6` derives the key as `sha256(4-digit PIN)` and uses `aes-256-ecb` (no IV, no auth tag). Keyspace is 10,000, so anyone who can read `settings.json` recovers the API key offline in milliseconds. The settings dir/file are created with default permissions (`settings/settings.ts:25,30`, i.e. 0644) and hold the key in plaintext when no PIN is set. The UI copy "API key is protected with PIN" oversells this.
   - Fix: `mkdirSync(..., { mode: 0o700 })`, `writeFile(..., { mode: 0o600 })`, switch to `aes-256-gcm` with a random 12-byte IV (store `iv:tag:ct`), and adjust the UI copy. Long-term: OS keychain.
2. **Path traversal via AE project-panel folder names.** `plainly-aescripts/src/utils.ts:62-70` (`getFolderPath`) joins raw `FolderItem.name` values; `collectProject.ts:48` and `copyFootage.ts:93-101` pass them to `path.join` + `mkdir` + `copyFile`. A folder named `..` (or `../../x`) in a shared `.aep` template writes footage outside `(Footage)`, and the undo only removes `(Footage)`.
   - Fix: sanitize each segment in `getFolderPath` (reject `.`/`..`, strip `/ \ : * ? " < > |`), and/or assert `path.normalize(itemAeFolder)` does not start with `..` on the Node side.
3. **Wrong-PIN false accept.** With ECB + PKCS#7, roughly 0.4% of wrong PINs decrypt "successfully" to garbage. `AuthProvider.tsx:45-50` then persists that PIN, every request 401s, and the PIN overlay never returns until panel restart. GCM tag verification fixes this as well.

## Correctness – high

[...]

PR #174, opened 2026-09-18
fix: keep footage with the same name from overwriting each other

vladavoX:
## Summary

Footage is copied into `(Footage)/<AE folder>/<file name>`, so two items sharing a basename inside the same After Effects folder raced for the same destination and both got relinked to whichever copy landed last. Colliding items now get a numbered subfolder next to the original one, and a sequence claims its whole frame numbering rather than only the frame the project points at.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Dependency update
- [ ] Docs / tooling

## Related issue

Refs #173 (10)

## Testing

- [x] Unit tests added / updated

Reproduced the overwrite manually and verified the fix manually.

**After Effects version tested:** AE 2026
**OS tested on:** macOS

## Screenshots

N/A, no UI change.

## Checklist


plainly-plugin/src/node/collectProject/utils.ts
vladavoX, line 1: should we write some tests confirming this works there also, even tho it works manually, we can cover more use-cases via tests?

vladavoX: Added in 20b8354 — 22 tests over two files.

`tests/collectProjectUtils.test.ts` covers `resolveFootageFolders` directly, it is pure. `tests/copyFootage.test.ts` runs `copyFootage` against real temp directories instead of a mocked fs, so it also covers the frame matching a sequence copy relies on: a padded name stays inside its own width (`shot_001.png` does not pull in `shot_0001.png`) while an unpadded one grows past it (`f9.png` -> `f10.png`), and the still/sequence deduplication keeps the sequence.

The case you would actually hit is covered end to end: two `logo.png` from different folders land in `Assets/` and `Assets/(2)/` with the right contents.

One thing worth knowing for future tests here, both `../src/node/bridge` and `../src/node/constants` build a `CSInterface` at module load, so both need mocking.

danixeee (approved):

